PR #1, opened 2019-09-07
Fix outdated Cell Representation definition

EmelyanenkoK:
In current codebase (see https://github.com/ton-blockchain/ton/blob/master/crypto/vm/cells.h#L42 ) depths of the reference cells are used to calculate Cell Representation.


Vodkashot28:
Thankyou

PR #13, opened 2019-09-10
Update Ed25519.cpp

NarakattiTirumpo:
constant time compare


NarakattiTirumpo:
If some compilers do compare all 32 bytes, we can't be sure that all memcmp implementations do the same.

NarakattiTirumpo:
This code is how djb implemented crypto_verify_32, https://nacl.cr.yp.to/install.html
I added 'not' operator because TON implementation uses 'true' as return of successful check.
Since this is a test of TON, it should be fine talking about it here. Also thanks for dislike. I'm fine if my request will not be accepted.

PR #14, opened 2019-09-11
Used /usr/bin/env fift instead of /usr/bin/fift

salemgolemugoo:


poma:
This can lead to problems on system that will parse `fift -s` as executable name and will fail to find it. See  [this answer](https://stackoverflow.com/questions/4303128/how-to-use-multiple-arguments-for-awk-with-a-shebang-i-e/4304187#4304187) for more detailed explanation.

Though I agree this is better than using a fixed binary location (or at least change hardcoded location to `/usr/local/bin/fift` as available for writing on more systems)

PR #18, opened 2019-09-16
secret serialization at exportKey et al. fixed

olegvg:
There was a bug in code of tonlib's interface specified in `tonlib_api.tl` as:
`key public_key:bytes secret:secureBytes = Key;`
and connected with absence of serialization/deserialization (base64_encode / base64_decode) of a `secret` at level of td-json wire protocol.
It affects all the users of `libtonlibjson` liteclient library except `tonlib-cli` where the bug is complementary to library's one. 
Fixed in the 'tonlib' and 'tonlib-cli' respectively.


ton-blockchain:
Thank you for the report.
The problem was in serialization, not in `KeyStorage.cpp`.
Should be fixed already.

Issue #30, opened 2019-09-24
Installation fails, TonlibConfig.cmake is missing


zevlg:
Installation fails with next error:

```
...
[100%] Built target json2tlo
Install the project...
-- Install configuration: ""
-- Up-to-date: /usr/local/lib/libtdutils.a
-- Up-to-date: /usr/local/lib/libtdactor.a
-- Up-to-date: /usr/local/lib/libtddb.a
-- Up-to-date: /usr/local/lib/libtonlibjson.so.0.5
-- Up-to-date: /usr/local/lib/libtonlibjson.so
-- Up-to-date: /usr/local/include/tonlib/tonlib_client_json.h
-- Up-to-date: /usr/local/include/tonlib/tonlibjson_export.h
-- Up-to-date: /usr/local/lib/cmake/Tonlib/TonlibTargets.cmake
-- Up-to-date: /usr/local/lib/cmake/Tonlib/TonlibTargets-noconfig.cmake
CMake Error at tonlib/cmake_install.cmake:96 (file):
  file INSTALL cannot find "/home/user/dev/ton/tonlib/TonlibConfig.cmake".
Call Stack (most recent call first):
  cmake_install.cmake:55 (include)


Makefile:128: recipe for target 'install' failed
make: *** [install] Error 1
```

logvik:
Related to #35

kirelagin:
> Related to #35

No, not related.

PR #51, opened 2019-09-28
Fix `make install`

kirelagin:
* Add missing library config (`TonlibConfig.cmake`) using a standard template
* Make `make install` install binaries


kirelagin:
Fixes #30.

EmelyanenkoK:
Should be fixed now.

PR #55, opened 2019-09-29
Fixes for Fift semantics

kirelagin:
There are a couple of places where the semantics of Fift diverge from the specification. It seems that in all cases we found so far the specification was right, meaning that the behaviour described in it made more sense than what the implementation did, thus we fixed the implementation.

Too sad there is no easily usable testing framework, otherwise I would write defect test-cases.

### DIFF
--- a/crypto/fift/words.cpp
+++ b/crypto/fift/words.cpp
@@ -937,8 +937,9 @@ void interpret_fetch(vm::Stack& stack, int mode) {
     if (mode & 2) {
       stack.push(std::move(cs));
     }
-    stack.push_bool(false);
-    if (!(mode & 4)) {
+    if (mode & 4) {
+      stack.push_bool(false);
+    } else {
       throw IntError{"end of data while reading integer from cell"};
     }
   } else {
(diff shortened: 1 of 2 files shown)


ton-blockchain:
1) You are right about B@; however, there is a second place in this function to be fixed.
2) Current implementation of ref@ appears to conform to the specification:
  ref@+ (s – s' c), similar to ref@, but returns the remainder of s as well.
(first push modified Slice s', then the Cell c)
However, there is another bug in this function.
3) It seems that the semantics of "hash" and "Bhash" has been changed without updating the specification. Now these words are used in too many places to be changed back. We'll add two new words and update the specification.

kirelagin:
> Current implementation of ref@ appears to conform to the specification
Ok, you are right, it means that the specification is wrong too, because all other deserialisation functions leave the slice on top so that it’s easy to parse it further.

kirelagin:
Ok, I see you have made some changes. I have removed the `hash` commit and force-pushed the branch, but that’s it for now, I’ll resolve the conflicts later.

PR #58, opened 2019-09-30
Fixes for Fift semantics (2)

kirelagin:
This is a reincarnation of #55 with the remaining fixes that haven’t made it upstream yet.

### DIFF
--- a/crypto/fift/words.cpp
+++ b/crypto/fift/words.cpp
@@ -942,7 +942,9 @@ void interpret_fetch(vm::Stack& stack, int mode) {
   auto n = stack.pop_smallint_range(256 + (mode & 1));
   auto cs = stack.pop_cellslice();
   if (!cs->have(n)) {
-    stack.push(std::move(cs));
+    if (mode & 2) {
+      stack.push(std::move(cs));
+    }
     stack.push_bool(false);
     if (!(mode & 4)) {
       throw IntError{"end of data while reading integer from cell"};
--- a/crypto/fift/words.cpp
+++ b/crypto/fift/words.cpp
@@ -1030,14 +1030,14 @@ void interpret_fetch_ref(vm::Stack& stack, int mode) {
     }
   } else {
     auto cell = (mode & 2) ? cs.write().fetch_ref() : cs->prefetch_ref();
-    if (mode & 2) {
-      stack.push(std::move(cs));
-    }
     if (mode & 1) {
       Ref<vm::CellSlice> new_cs{true, vm::NoVm(), std::move(cell)};
       stack.push(std::move(new_cs));
     } else {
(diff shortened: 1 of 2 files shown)


ton-blockchain:
The idea is that after fetching a Cell we are likely to begin parsing it with a new <s, so it has to be on top; for other data types, it might make sense to leave the deserialized value under the top, and the remainder of the Slice on top to continue parsing.

For instance, <s 32 i@+ 32 i@+ ref@+ <s 256 u@+ s> swap 64 B@+ s>
could be used to parse
```
test$_ a:int32 b:int32 ^[ addr:uint256 ] signature:bits512 = Test;
```
as follows:
```
x{1234567887654321CD4D08C28F9BB0DB071596408E9BFC4F224AFE1A095834EE104AFEE62C1363FC3A50B2AA2C8BEC373F3FB8ACFDBE7C3DDBA01A1747C5D573BDCA55F5917E1173}
 x{A526955C4B1B3CC5D76B037AFCCB908C40FD19AA667EF613C70DEF9BD61EB0DE} |_
 ok
dup csr. s>c
x{1234567887654321CD4D08C28F9BB0DB071596408E9BFC4F224AFE1A095834EE104AFEE62C1363FC3A50B2AA2C8BEC373F3FB8ACFDBE7C3DDBA01A1747C5D573BDCA55F5917E1173}
 x{A526955C4B1B3CC5D76B037AFCCB908C40FD19AA667EF613C70DEF9BD61EB0DE}
 ok
<s 32 i@+ 32 i@+ ref@+ <s 256 u@+ s> swap 64 B@+ s> .s
305419896 -2023406815 74699791053130924547043279704806878252800984766281970511501321307481257717982 BYTES:CD4D08C28F9BB0DB071596408E9BFC4F224AFE1A095834EE104AFEE62C1363FC3A50B2AA2C8BEC373F3FB8ACFDBE7C3DDBA01A1747C5D573BDCA55F5917E1173 

```

ton-blockchain:
However, the second suggested commit 6ee05c5 is a valid bugfix anyway. Thanks!

kirelagin:
> after fetching a Cell we are likely to begin parsing it with a new <s

Ok, I see your point. I find it hard to quantify what is more likely: to begin parsing it or to unpack everything first and begin parsing the cell after that. From my (extremely limited) experience it might not be the case that it is “likely”, and, furthermore, there will be people who want to do it one way or the other, and in any case some of them will have to `swap`, so the question boils down to whether it is worth it to save one `swap` for some people at the cost of having the behaviour different from all other parsing operators.

Ultimately it is y’all who make this decision, I just wanted to bring this up, until there is still room for change, as, at the moment, there are no uses of `ref@` in the code base so it can be changed relatively cheaply.

PR #60, opened 2019-09-30
Do not push zero on stack in runvm

kelvich:
Fift has several words for running vm, but only `runvmcode` does not push zero on stack before run, which is handy for testing contracts with method_id-based get functions. However  `runvmcode` does not initialise `c3`/`c4` registers, so it is hard to test contracts with persistent storage.

For simple contracts it is possible to use `runvm` and change entry function by prepending small code:
```
<{ DROP 128517 PUSHINT }>s
"contract.fif" include <s
|_
```
But that method doesn't play well with `CALLDICT`.

So this patch removes automatic zero push in `runvm` and gives user ability to push method_id before run.


ton-blockchain:
We have decided to make the proposed change and now `runvm` will not push the implicit zero. In order to simplify local testing of smart contracts, we have also added new words `runvmctx` and `gasrunvmctx`in commit 29deff15c3cfbdfdb73cae37cde0831e9f079e01 that accept an extra value to initialize `c7` (the "blockchain context" register).

PR #71, opened 2019-10-03
fix return value

negasus:
Fix possible error with returns -1 instead false


ton-blockchain:
Already fixed in commit dd74548.

Issue #74, opened 2019-10-03
We've found errors in your code.


cerg2010cerg2010:
Hello! We've checked your project with PVS-Studio static analyzer and found some bugs in your code. [PVS-Studio](https://www.viva64.com/en/pvs-studio/) is a static code analyzer for C, C++, C#, and Java. You can read about errors in this article [here](https://www.viva64.com/en/b/0674/).

Also you can get a free license for one month as a gift:
```
Name: Telegram Open Network
Key: X95G-43MC-1KTG-T5G8
License Type: Team License
License Valid Thru: Nov 03, 2019
```

ton-blockchain:
Thanks! We'll fix almost all of these error in the next commit.

ton-blockchain:
Mostly fixed in commit dd745485e26cf581019c2b8c3013d4c14ec0473e

PR #75, opened 2019-10-03
fix less or equal builtin

cerg2010cerg2010:
See #74


ton-blockchain:
Fixed in dd745485e26cf581019c2b8c3013d4c14ec0473e

PR #78, opened 2019-10-03
Remove unused variable

Enuvid:
Deleted unused variable in the `set_var_info_except(const VarDescrList& new_var_info, const std::vector<var_idx_t>& var_list)`


ton-blockchain:
No, there was a bug in that function, but this fix is a bad idea.

PR #87, opened 2019-10-04
Add unsigned dictionary manipulation funC primitives

danog:
Unsigned keys in dictionaries are common when working, for example, with keys that are hashes (for example `HASH` returns an unsigned 256-bit integer; I guess signed keys would work just fine, but it's nicer this way).

Another set of primitives that I'm really feeling the lack of (both in TVM and in funC) are dictionary count primitives: a .count() method would be extremely nice, instead of being forced to use a loop to count elements; tuples already have such a primitive, but dictionaries are extremely handy when working with large keys and values, since TVM automatically handles serialization of large keys and values with references to separate cells (and now that I think of it, there's no native way to serialize a simple tuple to a HashMapE).

### DIFF
--- a/crypto/smartcont/stdlib.fc
+++ b/crypto/smartcont/stdlib.fc
@@ -63,9 +63,13 @@ tuple parse_addr(slice s) asm "PARSEMSGADDR";
 
 cell idict_set_ref(cell dict, int key_len, int index, cell value) asm(value index dict key_len) "DICTISETREF";
 (cell, ()) ~idict_set_ref(cell dict, int key_len, int index, cell value) asm(value index dict key_len) "DICTISETREF";
+cell udict_set_ref(cell dict, int key_len, int index, cell value) asm(value index dict key_len) "DICTUSETREF";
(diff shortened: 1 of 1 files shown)


ton-blockchain:
The idea was to extend `stdlib.fc` with interfaces to new TVM primitives whenever they are needed by one of the smart contracts instead of providing interfaces to all of 600+ TVM primitives from the very beginning. However, you have a good point here: at least unsigned integer / signed integer dictionary operations should be added simultaneously. We'll add some missing funC functions (a bit more than you suggest) in a future commit.

As to the count() primitive, the structure of an (unaugmented) dictionary does not admit an efficient way of implementing such a method. If you really need it, you could simply store the counter along with the dictionary and increment or decrement it when appropriate.

danog:
Great, thanks for the tip!

ton-blockchain:
Partially "fixed" in commit 29deff15c3cfbdfdb73cae37cde0831e9f079e01

PR #98, opened 2019-10-06
Fix shebang and permissions of fift scripts

danog:


ton-blockchain:
That's an interesting suggestion, but it does not work with GNU Coreutils 8.28. You probably have a FreeBSD or MacOS system.

danog:
Yeah I did have some doubts on this, it didn't seem particularly portable.
The best option would be just to create a symlink to the fift binary in /usr/bin/fift and just load the binary directly from there without having to rely on env, but I guess this is a minor issue.

PR #132, opened 2019-10-12
Fix slice_at func method

JIoJIaJIu:


ton-blockchain:
Thanks! We have fixed this bug in commit b978e27b2f2ca6e4403843b9d6bf3f407bf63499.

PR #151, opened 2019-10-15
Split slice in load style operation

Enelar:
Neat method that saved few hours on debugging.


ton-blockchain:
How is this different from the already existing `load_bits` with the same signature (but more efficiently implemented)?

PR #170, opened 2019-10-25
Extended information about validators election participants

EmelyanenkoK:
Add new getmethod participant_list_extended() to elector_code which returns also `max_factor` and `addresses`


ton-blockchain:
A slightly modified version of the suggested get method has been added to the elector smart contract.

PR #198, opened 2019-11-26
Build FunC compiler as a static lib too

m-kus:
Basically it's just an additional cmake target. 
The only change in the code is conditional exclude of `main` in `func.cpp`.


EmelyanenkoK:
Time has shown that the change is not in demand. Can be reopened if needed.

Issue #210, opened 2019-12-10
Problem with TVM opcodes BALANCE & RANDSEED


gr8den:
Opcodes BALANCE & RANDSEED have incorrect GETPARAM index

Documentation (https://test.ton.org/tvm.pdf , A.11.4. Configuration primitives.) says
BALANCE equivalent to GETPARAM 6 and RANDSEED equivalent to GETPARAM 7
but in TL scheme
```
[ magic:0x076ef1ea actions:Integer msgs_sent:Integer
unixtime:Integer block_lt:Integer trans_lt:Integer
rand_seed:Integer balance_remaining:[Integer (Maybe Cell)]
myself:MsgAddressInt global_config:(Maybe Cell)
] = SmartContractInfo;
```
they have other indexes: RANDSEED is 6, BALANCE is 7

gr8den:
fix: https://github.com/ton-blockchain/ton/pull/211

ton-blockchain:
You are right. This will be fixed in the next commit, along with the new PRNG instructions.

ton-blockchain:
Should be fixed in commit d41ce55305718d207867f0dac60319ec4b6846f2. Please check.

PR #211, opened 2019-12-10
fix TVM opcodes BALANCE & RANDSEED

gr8den:
fixes issue https://github.com/ton-blockchain/ton/issues/210


ton-blockchain:
Fixed (in a slightly more correct way) in commit d41ce55305718d207867f0dac60319ec4b6846f2.

PR #215, opened 2019-12-15
get_account_address for raw_initialAccountState fix

koinov:
fix for tonlib client. state and data comes as strings, vm::std_boc_deserialize need slices.

### DIFF
--- a/tonlib/tonlib/TonlibClient.cpp
+++ b/tonlib/tonlib/TonlibClient.cpp
@@ -1058,8 +1058,8 @@ td::Result<block::PublicKey> get_public_key(td::Slice public_key) {
 }
 
 td::Result<block::StdAddress> get_account_address(const tonlib_api::raw_initialAccountState& raw_state) {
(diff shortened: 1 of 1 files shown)


ton-blockchain:
Why do you think this is necessary? `td::Slice` has an implicit constructor from `std::string`.

ton-blockchain:
We don't see what this commit is supposed to fix. Passing `std::string` as an argument to `vm::std_boc_deserialize` is perfectly valid; the argument will be automatically converted to `td::Slice`. Then `vm::std_boc_deserialize` will process this data without storing any long-term references to the data inside the string, so there is no need to create a new separate copy of the data. (In fact, if a function creates long-term references to the data it receives as an argument, type `td::BufferSlice` is used instead of `td::Slice`.)

koinov:
Hmm, I checked it now again and it perfectly works indeed. There's a chance my build was outdated, not sure now.

PR #218, opened 2019-12-21
Fix for not enough space for the wallet.fif message body slice

blackyblack:
body-cell slice bits are calculated before 1 i flag so there is a case when builder has exactly the body-cell length bits left and body-cell won't fit.


ton-blockchain:
Fixed in this and several similar wallet management scripts in a different fashion.

Issue #213, opened 2019-12-13
validator-engine-console -c and readline-related bug


Skydev0h:
Running **validator-engine-console** with **-c** option will cause a bug in console, all typed characters will be invisible (have to relogin to fix).
Nevertheless, it is possible to workaround this with **--disable-readline**.
lite-client works correctly for that matter.

Skydev0h:
Fixing this seems to be as simple as adding
`readline_enabled_ = false;` to `ValidatorEngineConsole::add_cmd`
It is present in corresponding `TestNode::add_cmd` function.

Skydev0h:
Added a commit fixing this in #220 PR.

Skydev0h:
Fixed in [this commit right here](https://github.com/ton-blockchain/ton/commit/acf16718e6385685828316ec116f12cb60455047#diff-44b6014c892833c2bce05d390166281cR88).

PR #220, opened 2019-12-22
FunC improvements (at the moment): unused x warnings, constants, include

Skydev0h:
Added (gas)**runvmctxact**(q) functions that output c5 state register (output actions) into the stack, allowing to inspect them for testing. gas-prefixed version also ouputs used gas and takes gas limit from input (like usual gas ones). q-suffixed version does not output every step when executing.

This addition was created to support advanced test suite for my [TON Blockchain Contest #2 (Freestyle contracts)](https://github.com/Skydev0h/ton-freestyle) and may be just useful for others :-).

Edit: Also fixed that run_vm_code ignored contract's commits before crashing.


Skydev0h:
Added a new commit that implements sdict functions that operate on slice keys.
Interoperability with VM DICTGET and DICTSET was verified in test.

Skydev0h:
Added a tiny miniscule but useful commit fixing #213.

Skydev0h:
Added a -d flag to func that would display warnings about all disabled (optimized out) function calls that would allow to find many different mistakes in code, such as incorrect usage of impure or pure functions, mistypes, accidental redefenitions, copy-paste mistakes, etc.

Skydev0h:
I like your way of thinking, the universal vm execution function is really useful due to its customization, and nice to see that you accounted for almost all ideas (and fix) except for the func optimization warnings.
They may be useful to combat [this "bug"](https://github.com/plexar88/ico/blob/master/bugs.txt).

Skydev0h:
Implemented basic support for integer constants: 96db6538.

They can be used like `const name = value;`
`value` must be a number or name of another constant.

The use of constant completely emulates use of it's value as close as possible.
This allows to maintain usability of code while not sacrificing optimizations (constant is treated as it's value in-place) and improving code readability (against pure numbers and even against asm or function-based constants, no need for braces).

In order to prevent confusion redefinition of constants is forbidden.

Caveat: expressions in value not yet supported (`const a = b + c;`).
Support of constants of other types is not useful, because constants are pseudo compile-time (their value must be injected while lexer is parsing), and no other simple types are supported.

Test case:
```
;; test constant stacking
const a = 111, b = 222;
const c = 123, d = c;
const e = 257; ;; large constant

;; const c = 124; ;; NOT permitted

int const_test(slice s, int v) {
	;; const f = 1; ;; NOT allowed, constants are global
	;; directly load constant into variable
	int z = a;
	;; use constant as native function argument
	int y = s~load_uint(b); ;; 222 LDU
	;; use variable as native function argument
	int x = s~load_uint(v); ;; ... LDUX
	;; test behaviour of direct and indirect const
	return x + y + z + c + e;
	;; will have at least 123 ADDCONST and 257 PUSHINT ADD
}

int const_test_bitwise(int v) {
	return v | a & b;
}

int const_test_condition(int v) {
	if (v == a) { return b; }
	elseif (v > c) { return e; }
	else { return c; }
}
```
generates code:
```
DECLPROC const_test
DECLPROC const_test_bitwise
DECLPROC const_test_condition
const_test PROC:<{
  SWAP
  222 LDU
  ROT
  LDUX
  DROP
  SWAP
  ADD
  111 ADDCONST
  123 ADDCONST
  257 PUSHINT
  ADD
}>
const_test_bitwise PROC:<{
  111 PUSHINT
  222 PUSHINT
  AND
  OR
}>
const_test_condition PROC:<{
  DUP
  111 EQINT
  IFJMP:<{
    DROP
    222 PUSHINT
  }>
  123 GTINT
  IFJMP:<{
    257 PUSHINT
  }>
  123 PUSHINT
}>
```

Skydev0h:
Added `include` keyword that allows to include another .fc files: 469fcd1.
Circular references are automatically protected from because while parsing source code func locks the source file from subsequent access, therefore if circular reference happens compilation would fail.
Keyword usage: `include "filename.fc";`

Skydev0h:
A **big update** to the unused variables and calls detection algorithm 67d3d21: separated **_Unused** variables and calls from **_Replaced** ones. The latter are replaced with another instructions or constants when builtins are executed on compilation. This way, they no more emit a warning, if really used.

Also, now **-u** flag is incremental like the -v, can be used once for warning only about unused variable assignment, and twice to warn also about unused function and operator calls.

Skydev0h:
_A neutron hit this PR and it was split into separate branches for maintainability:_
_#227 PR for constants support_
_#228 PR for include keyword_
_#229 PR for unused things detector_
_Another PR 230 was made, but it was not contained in this heavy nucleus._

**Closing this PR in favor of maintainable shardPRs.**
However, already implemented to now commits were not split into PRs for obvious reasons.
Those include (gas)vmrunctxact(q), commited states, sdict operations and vec fix.
_Never forget._

PR #227, opened 2020-01-05
FunC: Integer constants basic support

Skydev0h:

This PR implements `const` keyword that allows defining integer constants.

They can be used in code as like `const name = value;`
`value` must be a number or name of another constant.

The use of constant completely emulates use of it's value as close as possible.
This allows to maintain usability of code while not sacrificing optimizations (constant is treated as it's value in-place) and improving code readability (against pure numbers and even against asm or function-based constants, no need for braces).

In order to prevent confusion redefinition of constants is forbidden.

Caveat: expressions in value not yet supported (`const a = b + c;`). Such support requires precise and not so simple expression parsing, ensuring that only constants are used, immediately resolving them and applying operations while compiling. Not a task for the 4 AM brain.

Support of constants of other types is not useful, because constants are compile-time (their value must be injected while lexer is parsing), and no other simple types are supported.

P.S. I got really **determined** and finally understood FunC lexer and compiler logic after the last commit acf1671 and finally decided to actually implement something useful in FunC core.
_Too sad I did not get this enlightement 2-3 months ago._

Test case:
```
;; test constant stacking
const a = 111, b = 222;
const c = 123, d = c;
const e = 257; ;; large constant

;; const c = 124; ;; NOT permitted

int const_test(slice s, int v) {
	;; const f = 1; ;; NOT allowed, constants are global
	;; directly load constant into variable
	int z = a;
	;; use constant as native function argument
	int y = s~load_uint(b); ;; 222 LDU
	;; use variable as native function argument
	int x = s~load_uint(v); ;; ... LDUX
	;; test behaviour of direct and indirect const
	return x + y + z + c + e;
	;; will have at least 123 ADDCONST and 257 PUSHINT ADD
}

int const_test_bitwise(int v) {
	return v | a & b;
}

int const_test_condition(int v) {
	if (v == a) { return b; }
	elseif (v > c) { return e; }
	else { return c; }
}
```
generates code:
```
DECLPROC const_test
DECLPROC const_test_bitwise
DECLPROC const_test_condition
const_test PROC:<{
  SWAP
  222 LDU
  ROT
  LDUX
  DROP
  SWAP
  ADD
  111 ADDCONST
  123 ADDCONST
  257 PUSHINT
  ADD
}>
const_test_bitwise PROC:<{
  111 PUSHINT
  222 PUSHINT
  AND
  OR
}>
const_test_condition PROC:<{
  DUP
  111 EQINT
  IFJMP:<{
    DROP
    222 PUSHINT
  }>
  123 GTINT
  IFJMP:<{
    257 PUSHINT
  }>
  123 PUSHINT
}>
```

### DIFF
--- a/crypto/func/builtins.cpp
+++ b/crypto/func/builtins.cpp
@@ -27,7 +27,7 @@ using namespace std::literals::string_literals;
  * 
  */
 
-int glob_func_cnt, undef_func_cnt, glob_var_cnt;
+int glob_func_cnt, undef_func_cnt, glob_var_cnt, const_cnt;
 std::vector<SymDef*> glob_func, glob_vars;
 
 SymDef* predefine_builtin_func(std::string name, TypeExpr* func_type) {
--- a/crypto/func/func.h
+++ b/crypto/func/func.h
@@ -98,6 +98,7 @@ enum Keyword {
   _Asm,
   _Impure,
   _Global,
+  _Const,
   _Extern,
   _Inline,
   _InlineRef,
@@ -763,7 +764,18 @@ struct SymValGlobVar : sym::SymValBase {
   }
 };
 
-extern int glob_func_cnt, undef_func_cnt, glob_var_cnt;
+struct SymValConst : sym::SymValBase {
+  td::RefInt256 intval;
+  SymValConst(int idx, td::RefInt256 value)
+      : sym::SymValBase(_Const, idx), intval(value) {
+  }
+  ~SymValConst() override = default;
+  td::RefInt256 get_value() const {
+    return intval;
+  }
+};
+
+extern int glob_func_cnt, undef_func_cnt, glob_var_cnt, const_cnt;
 extern std::vector<SymDef*> glob_func, glob_vars;
 
 /*
--- a/crypto/func/keywords.cpp
+++ b/crypto/func/keywords.cpp
@@ -113,6 +113,7 @@ void define_keywords() {
       .add_keyword("forall", Kw::_Forall);
 
   sym::symbols.add_keyword("extern", Kw::_Extern)
+      .add_keyword("const", Kw::_Const)
       .add_keyword("global", Kw::_Global)
       .add_keyword("asm", Kw::_Asm)
       .add_keyword("impure", Kw::_Impure)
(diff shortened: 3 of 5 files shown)


ton-blockchain:
Nice and simple. Here are some thoughts on how more general constants of arbitrary types and constant expressions could be supported in FunC:
1) A good way to evaluate constant expressions would be to re-use the "abstract interpretation" machinery from `builtins.cpp`, otherwise all knowledge about evaluation of built-in functions (and operations) when applied to known constant arguments would have to be duplicated.
2) *Local* constants (declared inside functions and blocks) could be simply implemented by making `const` in such contexts almost equivalent to `var`, but setting a special "must_be_constant" flag for the defined variable, and checking that there are no assignments to this variable afterwards. After the optimizer finishes doing its forward passes, we check whether the RHS of each remaining assignment to a "must_be_constant" variable has a known constant value. The only drawback of this approach is that `const a = 2 + 3;` would not work with disabled optimizations. On the other hand, we might decide that local "constants" do not have a globally constant value, they are just variables that are assigned exactly once inside their scope. Then `int f(int x) { ... const a = x + 2; return a * a; }` would be always valid.
3) *Global* constants could be implemented similarly to what is suggested in this pull request, but with the following method of evaluating their values as arbitrary constant expressions:
3.1) The RHS *E* of a global assignment `const [type] x = E` is compiled as an anonymous function `_ anon_func() { return E; }` (such a function could be in fact created by the compiler, with a flag that prevents its emission into the output assembler code).
3.2) All type inference and optimizations are performed for this anonymous function, and then we check whether its `return` has a known constant value. If so, we add a constant symbol into the symbol table with this value and the inferred type.
4) This way of evaluating global constant expressions could be extended to support *lambda expressions* or *anonymous functions* in FunC. The syntax of a lambda could be `\args -> expr` or `\args { code ...; return value; }`, for instance, `\(int x) -> x * x`, `\x -> x * x` or `\(int x, int y) { x *= x; return x + y * 2; }`. It could be compiled as an anonymous global (inline_ref) function similarly to global constant expressions, and then a reference to this function could be pushed (by a variant of `PUSHCONT` or `PUSHCONTREF`) in the generated assembler code.
5) The only complication with such lambda expressions arises when they capture some of the variables defined outside their bodies; then a "closure" has to be generated: captured variables must be added as the first several arguments of the anonymous global functions, and suitable assembly code (using `SETCONTARGS`) would bind the code of the lambda with the actual values of the captured variables.

EmelyanenkoK:
Closed in favor to [func-constants](https://github.com/newton-blockchain/ton/tree/func-constants)

PR #228, opened 2020-01-05
FunC: Include keyword implementation

Skydev0h:

Added `include` keyword that allows to include another .fc files from the source.

Circular references are automatically protected from because while parsing source code func locks the source file from subsequent access, therefore if circular reference happens compilation would fail.
Relative paths to scripts are handled correctly as well (both in value and source relativity).

Keyword usage: `include "filename.fc";`

### DIFF
--- a/crypto/func/func.h
+++ b/crypto/func/func.h
@@ -35,6 +35,7 @@ namespace funC {
 
 extern int verbosity;
 extern bool op_rewrite_comments;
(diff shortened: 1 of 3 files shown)


EmelyanenkoK:
Merged in https://github.com/ton-blockchain/ton/pull/358

ferdragonf-spec (commented):
F

PR #229, opened 2020-01-05
FunC: Detection of unused variables (assigns) and (proc) calls

Skydev0h:

This massive (in terms of lines, to say the least) upgrade adds a counting `-u` flag that allows to warn about something really missing out: unused variables (assignments) if used at least once, adding unused procedure (and operator) calls if used at least twice.

If used, the compiler would display warnings about all unused things for this level (depending of number of `-u` flags used), that would allow to find many different mistakes in code, such as incorrect usage of impure or pure functions, mistypes, accidental redefenitions, copy-paste mistakes, etc.

Technically, replacements are marked with special _Replaced flag, while unused calls and variables only have the _Unused flag. This way, it is possbile to separate builtin-precalculated from really unused vars and calls.

As a room for improvement: I have not yet found a way to beautifully extract the name of unused variable (that should be relatively easy), ignore assignments to _ (holes) and correctly handle and display warnings about `(..., ..., ...) = (...~..., ...~..., ...~...)` functions (because of nature of tlide op).

### DIFF
--- a/crypto/func/abscode.cpp
+++ b/crypto/func/abscode.cpp
@@ -285,6 +285,7 @@ void Op::show(std::ostream& os, const std::vector<TmpVar>& vars, std::string pfx
     os << " ]\n";
   }
   std::string dis = disabled() ? "<disabled> " : "";
+  if (replaced()) dis = "<replaced> "; // not just disabled
   if (noreturn()) {
     dis += "<noret> ";
   }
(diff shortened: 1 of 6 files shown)


Skydev0h:
It may be worth to reconsider this once again since master was merged into the branch and merge conflicts were resolved.

Also the code was adapted for new subdivisions and precalculated operations.

zahrasohkandan-pixel:
waller sara

PR #230, opened 2020-01-05
Fift TVM assembler: unused procs optimizations and warnings

Skydev0h:
_Surprisingly, this PR is not related to 220, finally found a way not to dump into master_

This improvement on TVM assembler fift library is about detecting unused procedures (that have never been called), and removing them from resulting generated code. Optional warnings may be enabled about such situations and mixing inline and non-inline call of specific procedure.

If a procedure is called, it is marked (same for inline) and before generating the code unused procedures are filtered out from resulting code dictionary. With using some controlling methods the behavior of this algorithm can be altered:

`asm-retain-all` prevents removing unused procedures (old behavior)
`asm-warn-unused` prints warnings to stdout about each unused procedure
`asm-warn-inline-mix` prints warnings if function is called both inline and not

Those methods shall be invoked before running the fift assembler (calling PROGRAM{ ... ) however only after including Asm.fif to have effect (somewhere between them).

Smart algorithm detects if function is called only inline and optimizes it out without warning about it.
Moreover, procedures marked as method_id (automatic or explicit) and procedures with zero or negative numbers (message and special event handlers) are never optimized out and warned about.

### DIFF
--- a/crypto/fift/lib/Asm.fif
+++ b/crypto/fift/lib/Asm.fif
@@ -1065,11 +1065,18 @@ variable @proccnt
 variable @proclist
 variable @procdict
 variable @gvarcnt
+variable @usedproc
+variable @inlineproc
+variable @retainall
+variable @warnunused
+variable @warninlinemix
 19 constant @procdictkeylen
 { @proclist @ cons @proclist ! } : @proclistadd
 { dup @procdictkeylen fits not abort"procedure index out of range"
   1 'nop does swap dup @proclistadd 0 (create)
 } : @declproc
+{ <b b> <s swap @usedproc @ @procdictkeylen idict! drop @usedproc ! } : @useproc
+{ <b b> <s swap @inlineproc @ @procdictkeylen idict! drop @inlineproc ! } : @inlproc
 { 1 'nop does swap 0 (create) } : @declglobvar
 { @proccnt @ 1+ dup @proccnt ! @declproc } : @newproc
 { @gvarcnt @ 1+ dup @gvarcnt ! @declglobvar } : @newglobvar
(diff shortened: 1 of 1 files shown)


ton-blockchain:
This pull request is also a very nice and at the same time a very compact one. The only minor drawback that we could find is that it would make `CALL` / `JMP` / `PREPARE` work in a slightly strange fashion outside the context of a large program, for example, if one just wants to compile `<{ 239 INT 17 CALLDICT }>s` separately (maybe to be used as a trampoline).

That said, we'll think about integrating these changes, maybe with minor modifications.

Skydev0h:
@ton-blockchain I do not see problems with use of those words outside of context of program.
The aforementioned operation compiles fine to `x{8100EFF011}`.

Invocation of `@useproc` should cause no harm ouside of context of program. It would add an empty entry to `@usedproc` that would not be used because that dict is used only in `}END` construction.

It also would not have side effects to large program if it is built later because `PROGRAM{` resets ~all hopes and dreams back to zero~ `@usedproc` back to `dictnew`.

P.S.: I just checked, it actually leaves residual key in `@usedproc` dictionary, it would contain the entry with key 17: `x{A600046_}`

Skydev0h:
It seems that something like this [was implemented here](https://github.com/ton-blockchain/ton/commit/1de39f5d7cc1655ae691462370dffec26cd9bc77#diff-5f03b33ac2b31f60a39c7b33ec860850), but I am still trying to understand what is going on there.
Sometimes, fift code is hard to understand right away :)

ton-blockchain:
You are right. This is a partial implementation of the feature that we wanted to test first (removal of inlined functions from the dictionary) for the manual DNS smart contract, along with some other not so obvious but useful optimizations. The two other features (optional warnings about unused procedures and mixed inline/called procedures) can be now implemented very easily, in two extra lines of Fift code.

As a result, BoC-serialized code of the manual DNS smartcontract became 666 bytes long instead of 732.

ton-blockchain:
Implemented in commit 493ae2410cda47dffc7e793d5db28fa84c4f88a5. Please check.

Skydev0h:
Yeah, noticed that. Very nice one! A good basement for even more spaghetti-less logic.

PR #231, opened 2020-01-09
Fift: Add a flag to support backtracing error call stack

Skydev0h:
This PR implements a `-t` flag that activates error tracing mode.

That would cause detailed backtrace be printed as WARN and INFO in cause of error during compilation or interpretation (execution) of a fift script.
This can help debugging errors tremendously, because otherwise only top-level called word is displayed in error, and it may be **very very** difficult to find out where error has occured.
Now debugging such errors becomes several orders of magnitude easier.

If the flag is activated and error occurs, much information would be printed to 2 (WARN) and 3 (INFO) levels during unwinding including call stack of words, position of called words, currently executed branches of cond and while, operation mode of `funny loop`.

---

Technically additional operation on dictionary, reverse lookup, was implemented, to allow to (approximately) find the called word name when unwinding (reverse iterator is used to find out first defined instance of word - it should be most semantically correct).

Execution tokens are displayed as some hex number (picked to be most adequate, not very large but reasonably unique in execution), but currently I can't find the link between token address and the actual word list to be executed, it would be very expressive when this is fixed.

`cond` had to be rewritten into _ordinary word_ instead of _tail execution word_ because otherwise it is not possible to catch which branch was actually executed and display corresponding message during unwinding.

An example of "informative" error message without the -t flag:
image

And a much more **informative** backtrace with the -t flag:
image
Now the origin of error is perfectly traceable.

It may also be reasonable to snapshot stack before word execution and in case of error dump, but that is not implemented yet and may incur some more overhead, and add much mess to output.

### DIFF
--- a/crypto/fift/Dictionary.cpp
+++ b/crypto/fift/Dictionary.cpp
@@ -24,9 +24,30 @@ namespace fift {
 // WordDef
 //
 void WordDef::run(IntCtx& ctx) const {
-  auto next = run_tail(ctx);
-  while (next.not_null()) {
-    next = next->run_tail(ctx);
+  bool loop_provoked = false;
+  try {
+    auto next = run_tail(ctx);
+    while (next.not_null()) {
+      auto curr = next;
+      try {
+        next = next->run_tail(ctx);
+      } catch (...) {
+        if (ctx.tracing_errors()) {
+          LOG(INFO) << "Backtrace: "
+            << ctx.dictionary->reverse_lookup(curr.get()) 
+            << "(word execution)";
+          loop_provoked = true;
+        }
+        throw;
+      }
+    }
+  } catch (...) {
+    if (ctx.tracing_errors() && !loop_provoked) {
+      LOG(INFO) << "Backtrace: "
+        << ctx.dictionary->reverse_lookup(this) 
+        << "(word execution)";
+    }
+    throw;
   }
 }
 
@@ -78,7 +99,29 @@ Ref<WordDef> WordList::run_tail(IntCtx& ctx) const {
   }
   auto it = list.cbegin(), it2 = list.cend() - 1;
   while (it < it2) {
-    (*it)->run(ctx);
+    try {
+      (*it)->run(ctx);
+    } catch (...) {
+        if (ctx.tracing_errors()) {
+          std::ostringstream wl;
+          std::streampos offset = 0;
+          std::size_t offlen = 0;
+          auto iit = list.cbegin();
+          while (iit != list.cend()) {
+            if (iit == it) offset = wl.tellp();
+            std::string name = ctx.dictionary->reverse_lookup(iit->get());
+            wl << name;
+            if (iit == it) offlen = name.length() - 1;
+            ++iit;
+          }
+          auto current = ctx.dictionary->reverse_lookup(this);
+          auto spaces = std::string(current.size() + 4 + offset, ' ');
+          LOG(WARNING) << "Backtrace (word list execution):"
+            << "\n\t" << current << ": { " << wl.str() << "}"
+            << "\n\t" << spaces + std::string(std::max(offlen, 1ul), '^');
+        }
+        throw;
+    }
     ++it;
   }
   return *it;
@@ -134,6 +177,25 @@ WordRef* Dictionary::lookup(td::Slice name) {
     return nullptr;
   }
   return &it->second;
+}
+
+std::string Dictionary::reverse_lookup(const WordDef* ref) {
+  auto it = words_.crbegin();
+  while (it != words_.crend()) {
+    if (it->second.get_def().get() == ref)
+      return it->first;
+    ++it;
+  }
+  auto base_addr = (long long)(words_.crbegin()->second.get_def().get());
+  std::ostringstream os;
+  auto addr = (long long) ref;
+  //if (ref->inner_addr())
+  //  addr = ref->inner_addr();
+  os << "{" << std::hex 
+     << ((addr - base_addr) / 0x10 % 0x10000) 
+     << "} ";
+  std::string str = os.str();
+  return str;
 }
 
 void Dictionary::def_ctx_word(std::string name, CtxWordFunc func) {
--- a/crypto/fift/Dictionary.h
+++ b/crypto/fift/Dictionary.h
@@ -49,6 +49,9 @@ class WordDef : public td::CntObject {
   virtual const std::vector<Ref<WordDef>>* get_list() const {
     return nullptr;
   }
+  virtual long long inner_addr() const {
+    return 0;
+  }
 };
 
 class StackWord : public WordDef {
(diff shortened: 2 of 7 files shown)


Skydev0h:
Added that using flag second time (for example `-tt`) would dump stack on each unwinding frame **before** failed execution of the word.
Aforementioned example with `-tt` would now look like this:
image
That might help debugging for understanding developer.

ton-blockchain:
This pull request appears to be a nice implementation of a debug feature in Fift that different people needed before. You are really becoming an expert in Fift and FunC hacking. We'll think a bit whether the suggested changes could be integrated as is, or some modifications would be required beforehand, for example to improve efficiency. Some remarks:
1) Wouldn't it be more efficient to check `tracing_errors` at the very beginning of `WordDef::run()`, so that the execution would not be slower when the debug features are disabled?
2) The implementation of several words, such as `interpret_cond()` (the implementation of `cond`), has to be replaced with a less efficient one (disabling tail recursion optimizations and introducing more checks and more exception handling). Wouldn't it be more efficient to retain two versions of their implementations, say, `interpret_cond_debug()` and `interpret_cond_fast()`, and install one of them into the dictionary during startup (in `init_words_common()`) depending on whether `tracing_errors` is enabled?
3) The suggested implementation is not especially good at dumping constants embedded into Fift word definitions. We think a special virtual method (say, `dump(std::ostream&)`) could be added to `WordDef` (and specialized in derived classes) to improve dumping references to such execution tokens inside word lists.
4) The above optimizations seem quite straightforward. However, there is another less evident problem. We planned to improve the Fift interpreter by defining and implementing *Fift continuations* (eliminating all recursion from the Fift interpreter, thus transforming it into CPS). This would be useful, for instance, for a version of Fift integrated with the Lite Client. Special Fift words might be used to initiate LiteServer queries in such a version, and a continuation would have to be stored so that the execution would resume at correct point after receiving an answer, even if the Fift word was invoked not at the top level. Now we feel that accepting the suggested changes in Fift interpreter would be in conflict with the planned CPS transformation of the interpreter. It may happen that the best course of action would be to transform the interpreter to CPS first, and only then add the debug backtrace/stack dump features as suggested in this pull request.

Skydev0h:
@ton-blockchain Thank you for your review. I too think that some review and modifications should be made, I noticed that you always assess PRs as baseline and improve them with your plans in mind.
The main objective in this PR is the research of the idea of error backtrace, showing that it is actually possible. Without such feature it is tremendously difficult to find actual case of errors in many layers of nested calls.
1. It must be more efficient, this improvement was actually made iteratively, and is not absolutely finished (I could not find a way to beautifully link the word callers and called words themselves, those {xxxx} anons). Therefore it is a bit of mess and not throughtly optimized. Another thing in mind that I kept while doing it that way is that fift scripts currently are not prepared for high-speed execution, they are more like shell scripts where several milliseconds difference per command may not play role.
2. Very good idea. Did not think about conditional dictionary configuration. Will definitely keep this idea in mind, it may be useful to do some more intersting tricks.
3. Yeah, as I understand the problem is that constants are actually automatically executing "closures" that push required constant immediately. The extent to which backtracer was implemented should be enough to understand where exception happens looking at the dictionary words and assessing the offset (in number of words) from understood construction. Good point anyway.
4. Oh damn it. I was thinking about implementing such thing: Fift scripts, that can be called from Lite Client, and if invoked that way they could access blockchain (both read access by calling methods and write access by sending BoCs, if permitted by security policy). Good that I did not dive into that, because I did not even think about problem for storing the stack of continuations, because now it is really straightforward - a native stack of function calls that unwinds by itself with help of C++ call stack and automatic destruction (savepoints for includes). 
Now I feel a bit sad that I did not propose such idea earlier. _Looking at how data is pulled with lite client, feeded into scripts, and result of scripts is feeded into lite client the ability to execute scripts from lite client that could be able to interact with blockchain is obvious missing link in chain. I started thinking about it after inspecting the election bid creator in validator-engine that actually shows that it may be a good perfectly reasonable idea to call fift from the LC or VE._

Another idea that I have now but not implemented yet is interactive fift scripts that can request input from console. It may be good to make interactive console fift scripts, but it not so useful if fift considered as an utility script called by other programs (thats why my scripts in first contest had -r and -R mode for parseable output). But if scripts are executed from lite client it may be cool to be able to at least ask confirmation from user for some actions.

**I completely agree that it would be very reasonable to first do CPS transformation, and I think that the CPS may be much more fun to work with, and even may make accurate backtracing even easier.**

ton-blockchain:
Well, you are not completely right when you think that Fift is not interactive. Try to run the following program:

```
{ { 0 word (number) dup 1 <> } { ' 2drop if ."not an integer" cr } while drop } : readint
{ newkeypair drop 256 B>u@ 16 >> } : randi
{ randi 100 mod 1+ ."Try to guess my number (1-100)" cr } : start-game
{ start-game 0 word drop {
  ."Enter your guess: " readint over - dup } {
  0< { ."too low" } { ."too high" } cond cr } while ."you are right!" cr } : game
game

```

Skydev0h:
@ton-blockchain Does not work. Looking at it I do not understand how it would read input.

WSL
image

SSH
image

No difference, with -s or without.
Behaves like attempting to read number from EOFed file.

ton-blockchain:
Remove last line `game` from the file `game.fif`, and execute it separately:

```
$ cat > game.fif
{ { 0 word (number) dup 1 <> } { ' 2drop if ."not an integer" cr } while drop } : readint
{ newkeypair drop 256 B>u@ 16 >> } : randi
{ randi 100 mod 1+ ."Try to guess my number (1-100)" cr } : start-game
{ start-game 0 word drop {
  ."Enter your guess: " readint over - dup } {
  0< { ."too low" } { ."too high" } cond cr } while ."you are right!" cr } : game
$ fift -i game.fif
game
Try to guess my number (1-100)
Enter your guess: 34
too low
Enter your guess: 78
too high
...
```

This trick does not work indeed if `game` is invoked not from the console; this could be easily fixed if necessary. This example is just a proof of concept.

Skydev0h:
Oh, right, interesting mechanic 😃. Stupid of me not to think about _interactive_ mode.
Although that looks more like abusing interactive mode than actual in-out interaction.
image
Now i understand simplicity of the code and abuse of mechanics: it keeps reading next `word` (more precisely, a line, sep is \0, like until EOF) and then plays with that string (convert to number, check that type is 1 (integer), use that int).
Although I could not break-out from the input loop after it started and got nice features such as correct interpreting of 0x12 as number as well. The only issue of such mode is that user is required to enter command manually so that word would read "code" from stdin rather than from file and output of ` ok` after command. That way the user can enter anything, even builtin commands, instead of what to be expected and receive unpredictable results.
It is also very wacky having to understand that here `word` would read from script and work on it's contents, and here `word` would read from stdin and process user input. 

I thought about something like `readline` that would read line from stdin and return it, `readint` that would read a number, maybe behaving like usual grabbing of number from stdin, maybe `readkey` that would immediately read unbuffered key allowing for interesting behaviour (for example rewriting one-line menus with \r or entering passwords), etcetra. Now it is a bit harder to think such because there is an _interactive_ mode despite that it is pretty low-level (direct input of fift command is expected at least at first) and hacky (would get `word` double-meaning spaghetti without understanding where it would actually read from).

Issue #84, opened 2019-10-04
[bug] Invalid wallet address


igroman787:
Hello.
When creating a new wallet myWallet_001, an incorrect wallet address was received:
```
Creating new wallet in workchain 0 
Saved new private key to file myWallet_001.pk
StateInit: x{34_}
 x{FF0020DD2082014C97BA9730ED44D0D70B1FE0A4F260810200D71820D70B1FED44D0D31FD3FFD15112BAF2A122F901541044F910F2A2F80001D31F3120D74A96D307D402FB00DED1A4C8CB1FCBFFC9ED54}
 x{0000000021549100C863943A30D3DEDB5DB9C93ACA52650530D3F8009730A6A9C3BEBD0D}

new wallet address = 0:f34f14374680af6f3756bf35b22ec60fedf68cef6f031ebc9588adbf5a9b231 
(Saving address to file myWallet_001.addr)
Non-bounceable address (for init): 0QAPNPFDdGgK9vN1a_NbIuxg_t9ozvbwMevJWIrb9amyMe4t
Bounceable address (for later access): kQAPNPFDdGgK9vN1a_NbIuxg_t9ozvbwMevJWIrb9amyMbPo
signing message: x{00000000}

External message for initialization is x{88001E69E286E8D015EDE6EAD7E6B645D8C1FDBED19DEDE063D792B115B7EB5364621190D1573702E16DF75AB831FC0938C61C6159EDF85C836F9F32E6DD6063B521D9760554B0469D1D8D2BD2D55D4A489AF15AD28E6A3BC4F5CE88418F43046A05C0200000001_}
 x{FF0020DD2082014C97BA9730ED44D0D70B1FE0A4F260810200D71820D70B1FED44D0D31FD3FFD15112BAF2A122F901541044F910F2A2F80001D31F3120D74A96D307D402FB00DED1A4C8CB1FCBFFC9ED54}
 x{0000000021549100C863943A30D3DEDB5DB9C93ACA52650530D3F8009730A6A9C3BEBD0D}

[...]

PR #239, opened 2020-02-05
fix bug: Invalid wallet address

igroman787:
This bug has been described here: https://github.com/ton-blockchain/ton/issues/84
It was fix, but not in all places.


igroman787:
Not relevant. Fixed by this commit: 77842f9b637dd2efcd684a4668e9ff4a173449f8

PR #276, opened 2020-04-07
Add Freebsd to preprocessor of rlimit.cpp

sonofmom:
rlimit.cpp will not compile otherwise because of missing headers.
Compiles fine under amd64 FreeBSD 12.1-RELEASE with this fix.


sonofmom:
This PR has been superseded by previously merged and is not needed any longer. Closing.

PR #309, opened 2021-02-09
Aleksejmelman/fix

AleksejMelman:
This fixes the error "LNK2019: unresolved external symbol _GetProcessMemoryInfo@12 referenced in function void __cdecl PrintMemoryInfo(unsigned long) (?PrintMemoryInfo@@YAXK@Z)" when one tries to build windows desktop client.

Psapi.lib is already added to TARGETLIBS and tonlib compiles with -DPSAPI_VERSION=1 automatically, however this is not enough.

#pragma comment(lib, "psapi.lib") helps the linker to add the 'psapi.lib' library to the list of library dependencies, as if you had added it in the project properties at Linker->Input->Additional dependencies


AleksejMelman:
wrong repo

PR #317, opened 2021-10-15
Create cmake --build . --target rldp-http-proxy

arzpgh60:


EmelyanenkoK:
Corresponding changes should be made to Makefiles

PR #340, opened 2022-05-02
Rebase newton history to original repository

EmelyanenkoK:
Slightly cleaned history from partial commits.


validator/manager.cpp
OMIDNIK7, line 1: validator/manager.cpp

PR #361, opened 2022-05-12
Extended fift

OmicronTau:
Add better fift namespaces and construction, also fift-based disassembler from https://github.com/OmicronTau/ton/tree/disasm compatible with rebased history.


EmelyanenkoK:
There are minor changes into TVM (need to test them in testnet first?), also we probably need to update Fift docs prior to merging into main branches. Lets move to separate branch for now.

PR #394, opened 2022-06-06
Improve readme

butterfly1924:


EmelyanenkoK:
We prefer to keep plain Markdown format for now,

PR #395, opened 2022-06-06
Add translation for zh_CN

butterfly1924:


EmelyanenkoK:
We intentionally decided to keep all docs and information in English to prevent geographical skew and build TON as global community.
While Chinese, Russian, Iranian and other developers are important for us, at least for now, we will not make translations in ton-blockchain repos.

EmelyanenkoK:
We intentionally decided to keep all docs and information in English to prevent geographical skew and build TON as global community.
While Chinese, Russian, Iranian and other developers are important for us, at least for now, we will not make translations in ton-blockchain repos.

PR #385, opened 2022-05-28
add m1 support

awesome-doge:
Tested on Apple M1 Max
```
$ make -j 10          
                                                                                                                                                                    
[  0%] Built target crc32c_sse42
[  0%] Built target crc32c_arm64
[  0%] Checking the git repository for changes...
[  0%] Built target absl_dynamic_annotations
[  0%] Built target absl_log_severity
[  1%] Built target absl_exponential_biased
[  1%] Built target absl_spinlock_wait
[  2%] Built target absl_int128
[  4%] Built target absl_civil_time
[  4%] Built target absl_time_zone
[  4%] Built target absl_city
[  4%] Built target absl_raw_logging_internal
[  5%] Built target tdtl
[  5%] Built target crc32c
[  5%] Built target absl_debugging_internal
[  5%] Built target absl_strings_internal
[  5%] Built target absl_bad_optional_access
[  5%] Built target absl_throw_delegate
[  5%] Built target tonlib_generate_java_api
[  5%] Built target absl_bad_variant_access
[  5%] Built target absl_base
[  5%] Built target absl_stacktrace
[  5%] Built target absl_demangle_internal
[  5%] Built target absl_malloc_internal
[  6%] Built target absl_strings
[  6%] Built target absl_graphcycles_internal
[  8%] Built target absl_symbolize
[  8%] Built target absl_hash
[  9%] Built target absl_time
[  9%] Built target check_git
[  9%] Built target absl_synchronization
[  9%] Built target git
[  9%] Built target absl_hashtablez_sampler
[  9%] Built target absl_raw_hash_set
[ 15%] Built target tdutils
[ 15%] Built target memprof
[ 16%] Built target tdactor
[ 16%] Built target generate_common
[ 17%] Built target tdfec
[ 17%] Built target tl_generate_common
[ 17%] Built target benchmark
[ 17%] Built target tdnet
[ 17%] Built target terminal
[ 19%] Built target test-tdactor
[ 20%] Built target tddb_utils
[ 20%] Built target benchmark-fec
[ 20%] Built target test-fec
[ 20%] Built target tl_api
[ 23%] Built target test-tdutils
[ 24%] Built target tl_tonlib_api
[ 24%] Built target tl_lite_api
[ 24%] Built target tcp_ping_pong
[ 24%] Built target udp_ping_pong
[ 24%] Built target test-net
[ 24%] Built target fec
[ 24%] Built target tl_tonlib_api_json
[ 28%] Built target ton_crypto
[...]

PR #400, opened 2022-06-10
apple m1 support

tolya-yanot:
By @doronaviguy

Old PR - https://github.com/newton-blockchain/ton/pull/59

This PR includes the following

1. detecting arm based apple devices and passing the right value to clang complier
2. in case of ARM processor we are declaring the cpp version to be 14 set(CMAKE_CXX_STANDARD 14)
3. changing the clang flag from march to mcpu and pass the right flag based on the system architecture, in case of an apple arm based processor we are passing -mcpu=apple-m14. 

Tested on apple both arm and non arm cpus


EmelyanenkoK:
Close in favor #385

PR #401, opened 2022-06-11
Add apple m1 support

EmelyanenkoK:

PR #444, opened 2022-08-20
Compiler -march fix for not supported arch

Trinketer22:
I think there is an issue with the way -march flag is handled in the
current build system.

It is set to native (heavy cpu specific optimizations) via *TON_ARCH* and it is never
checked if compiler actually support specified arch.

That's what is causing all those issues with apple m1 builds or any
other arm system and not lack of -mcpu flag.

Without checking early, build will only fail at link stage.
Strict arch support checking is due.

1. Check if "-march=${TON_ARCH}" flag is supported by the compiler
   Adds it to *CXX_FLAGS* if all good

2. Otherwise if such flag is not supported but TON_ARCH is set to default we
   continue execution without setting march and leaving arch decisions to
   cmake build system

3. Otherwise user specified TON_ARCH intentionally and current arch is
   not supported by the compiler.
   Therefore we terminate build process immediately and let user know
   what happened.


Trinketer22:
#401 This one solves m1 issue not by setting *-mcpu* but by not setting *-march* which is a subset of what this pull is doing.
Also only most recent clang supports -mcpu=apple-m1.
I don't know if we should do anything about it but just saying.

PR #452, opened 2022-09-06
Add ~strdump and STRDUMP TVM OP

tvorogme:
Now you can do: 

latest-screenshot

I've added realization of `FE14 — STRDUMP, dumps the Slice at s0 as an UTF-8 string.`

### DIFF
--- a/crypto/func/builtins.cpp
+++ b/crypto/func/builtins.cpp
@@ -1063,6 +1063,8 @@ void define_builtins() {
                       AsmOp::Nop());
   define_builtin_func("~dump", TypeExpr::new_forall({X}, TypeExpr::new_map(X, TypeExpr::new_tensor({X, Unit}))),
                       AsmOp::Custom("s0 DUMP", 1, 1), true);
+  define_builtin_func("~strdump", TypeExpr::new_forall({X}, TypeExpr::new_map(X, TypeExpr::new_tensor({X, Unit}))),
+                      AsmOp::Custom("STRDUMP", 1, 1), true);
   define_builtin_func("run_method0", TypeExpr::new_map(Int, Unit),
                       [](auto a, auto b, auto c) { return compile_run_method(a, b, c, 0, false); }, true);
   define_builtin_func("run_method1", TypeExpr::new_forall({X}, TypeExpr::new_map(TypeExpr::new_tensor({Int, X}), Unit)),
--- a/crypto/vm/debugops.cpp
+++ b/crypto/vm/debugops.cpp
@@ -105,6 +105,37 @@ int exec_dump_value(VmState* st, unsigned arg) {
   return 0;
 }
 
+int exec_dump_string(VmState* st) {
+  VM_LOG(st) << "execute STRDUMP";
+  if (!vm_debug_enabled) {
+    return 0;
+  }
+
+  Stack& stack = st->get_stack();
+  auto item = stack[0];
+
+  if (item.is(4)) {  // wanted t_slice
+    auto cs = item.as_slice();
+    auto size = cs->size();
+
+    if (size % 8 == 0) {
+      auto cnt = size / 8;
+
+      unsigned char tmp[128];
+      cs.write().fetch_bytes(tmp, cnt);
+      std::string s{tmp, tmp + cnt};
+
+      std::cerr << "#DEBUG#: " << s;
+      std::cerr << std::endl;
+    } else {
+      std::cerr << "#DEBUG#: slice contains not valid bits count";
+    }
+  } else {
+    std::cerr << "#DEBUG#: is not a slice" << std::endl;
+  }
+  return 0;
+}
+
 void register_debug_ops(OpcodeTable& cp0) {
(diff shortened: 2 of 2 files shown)


EmelyanenkoK:
May be it will be better to merge to SpyCheese/toncli-local branch?

tvorogme:
@EmelyanenkoK do you think it's not main repo related? Those TVM OPs defined in TVM white paper. I also want to add other debug OPs support as it defined in white paper because it's really hard to debug code without them. 

Also it's only debug TVM OPs, so there is no security issues on TON update, because those OPs (`0xfe14`) just silenced in production mode of TVM [here](https://github.com/ton-blockchain/ton/blob/0fc54b7749d40e94b0d7e2895b3238d25e7720e2/crypto/vm/debugops.cpp#L142). 

Also I think that this syntax will help our javascript-adepts :)

But if you think that new debug OPs better to store in `SpyCheese/toncli-local` - ok, will move PRs there.

EmelyanenkoK:
From one hand I agree that it should be ok to add to main repo since it is debug ops, from the other changing anything in core node without strict necessity looks frightening :)

tvorogme:
Sure. In any case, the final decision is up to TF Core, so just let me know what you come up with.To know where to do PR :)

crypto/vm/debugops.cpp
```diff
@@ -113,7 +144,9 @@ void register_debug_ops(OpcodeTable& cp0) {
   } else {
     // NB: all non-redefined opcodes in fe00..feff should be redirected to dummy debug definitions
     cp0.insert(OpcodeInstr::mksimple(0xfe00, 16, "DUMPSTK", exec_dump_stack))
-        .insert(OpcodeInstr::mkfixedrange(0xfe01, 0xfe20, 16, 8, instr::dump_1c_and(0xff, "DEBUG "), exec_dummy_debug))
+        .insert(OpcodeInstr::mkfixedrange(0xfe01, 0xfe13, 16, 8, instr::dump_1c_and(0xff, "DEBUG "), exec_dummy_debug))
```
SpyCheese: Max opcode is not inclusive, so it should be 0xfe14, not 0xfe13.

crypto/vm/debugops.cpp
```diff
@@ -105,6 +105,37 @@ int exec_dump_value(VmState* st, unsigned arg) {
...
 }
 
+int exec_dump_string(VmState* st) {
+  VM_LOG(st) << "execute STRDUMP";
+  if (!vm_debug_enabled) {
+    return 0;
+  }
+
+  Stack& stack = st->get_stack();
+  auto item = stack[0];
```
SpyCheese: Check depth before accessing the stack.
Don't use magic number (`.is(4)`). `.as_slice()` returns null if item is not a slice, so it can be used here.

crypto/vm/debugops.cpp
```diff
@@ -105,6 +105,37 @@ int exec_dump_value(VmState* st, unsigned arg) {
...
+      auto cnt = size / 8;
+
+      unsigned char tmp[128];
+      cs.write().fetch_bytes(tmp, cnt);
+      std::string s{tmp, tmp + cnt};
+
+      std::cerr << "#DEBUG#: " << s;
+      std::cerr << std::endl;
+    } else {
+      std::cerr << "#DEBUG#: slice contains not valid bits count";
```
SpyCheese: Missing std::endl

crypto/func/builtins.cpp
```diff
@@ -1063,6 +1063,8 @@ void define_builtins() {
                       AsmOp::Nop());
   define_builtin_func("~dump", TypeExpr::new_forall({X}, TypeExpr::new_map(X, TypeExpr::new_tensor({X, Unit}))),
                       AsmOp::Custom("s0 DUMP", 1, 1), true);
+  define_builtin_func("~strdump", TypeExpr::new_forall({X}, TypeExpr::new_map(X, TypeExpr::new_tensor({X, Unit}))),
+                      AsmOp::Custom("STRDUMP", 1, 1), true);
```
SpyCheese, line 1067: Please add "STRDUMP" to Asm.fif

tvorogme: https://github.com/ton-blockchain/ton/blob/35d17249e6b54d67a5781ebf26e4ee98e56c1e50/crypto/fift/lib/Asm.fif#L1170 since those debug OPs are defined in white paper they already defined in `Asm.fif`, but just don't have realization. 

Thanks for review, I'll fix all shortly.

tvorogme:
@SpyCheese all done. + little test:

```
"Asm.fif" include

// Not slice
<{ 0 PUSHINT STRDUMP }>s runvmcode 2drop
 
// Stack empty
<{ STRDUMP }>s runvmcode

// Not valid bits count
<{ b{0101010} PUSHSLICE STRDUMP }>s runvmcode 2drop

// All work as expected!
<{ "INFO: 🍿 All work fine" $>s PUSHSLICE STRDUMP }>s runvmcode 2drop
```

PR #453, opened 2022-09-07
Add parameter count to tonlib method raw.getTransactions

dungeon-master-666:
Currently, there's a hardcoded value `count=10`. This PR adds `count` to tonlib interface. To keep backward compatibility if parameter is not present it's set to default value 10.


dungeon-master-666:
Closing this PR in favor  #461

PR #455, opened 2022-09-09
Func and Fift lib for WASM

AlexeyFSL:
FunC compiler refactored. Some logic has been moved from main() to funC::func_proceed function. The main function has been moved to func-main.cpp file.
Output is now written to any std::ostream instead of std::cout & std::cerr.

Added new library (funcfiftlib) as executable target for compiling FunC smart-contracts straight to BOC packed in JSON string. 
This executable uses both FunC compiler & Fift interpreter internally. 

Main purpose of funcfiftlib is to compile to WASM via Emscripten to be used in JS packages. Also this library needs some extended libraries (openssl and zlib) compiled by Emscripten too. Fiftlib files from crypto/fift/lib/ was included into WASM.
The build script added into .github/scripts/ folder.


SpyCheese:
`std::cerr` was changed to a custom std::ostream, but only in func.cpp. There's output to std::cerr in other files, too, and that output will not be captured by ostream in func_proceed

Perhaps this is not important, because custom err stream is used only in funcfiftlib.cpp, and it sets verbosity to 0.

Wyeth8:
Run

Shayan1191 (commented):
Ok

PR #486, opened 2022-10-10
Add account state by transaction and emulator

EmelyanenkoK:


tonlib/tonlib/TonlibClient.cpp
```diff
@@ -1633,6 +1660,302 @@ class GetShardBlockProof : public td::actor::Actor {
...
+  td::unique_ptr<AccountState> account_state_;
+  std::vector<td::Ref<vm::Cell>> transactions_; // std::vector<^Transaction>
+
+  size_t count_{0};
+  size_t count_transactions_{0};
+  bool incomplete_{true};
+  bool stopped_{false};
+
+  void get_block_id(td::Promise<FullBlockId>&& promise) {
+    auto query = ton::lite_api::liteServer_lookupBlock(0b111111010, ton::create_tl_lite_block_id_simple({request_.address.workchain, ton::shardIdAll, 0}), request_.lt, 0);
```
SpyCheese, line 1701: 1) Here we need to find a block with the account. Instead of `ton::shardIdAll`, use 60-bit prefix of account address (liteserver will correctly handle that request).
2) To get the correct previous block, use the corresponding bit of account address (not `ton::is_left_child(block_id.id.shard)`).

emulator/transaction-emulator.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+  if (!trans->compute_phase->accepted) {
+    if (external) {
+      // inbound external message was not accepted
+      auto const& cp = *trans->compute_phase;
+      return td::Status::Error(
+          -701,
+          PSLICE() << "inbound external message rejected by transaction " << acc->addr.to_hex() << ":\n" <<
+              "exitcode=" << cp.exit_code << ", steps=" << cp.vm_steps << ", gas_used=" << cp.gas_used <<
+              (cp.vm_log.empty() ? "" : "\nVM Log (truncated):\n..." + cp.vm_log));
```
SpyCheese, line 200: `cp.vm_log` will be empty by default.
You can add `compute_phase_cfg.with_vm_log = true;` to enable it.

EmelyanenkoK:
Closed in favor of #592

PR #499, opened 2022-10-24
Fix invalid numeric argument '/Wextra' when compile on Windows

neodiX42:
Execution of targets:

```
cmake --build . --target prepare_cross_compiling
cmake --build . --target native-lib
```
inside example\android, fails on Windows with error:
cl : Command line error D8021 : invalid numeric argument '/Wextra'.

Also makes strip command compatible with Darwin OS


Richard9012:
> Execution of targets:
> 
> ```
> cmake --build . --target prepare_cross_compiling
> cmake --build . --target native-lib
> ```
> 
> inside example\android, fails on Windows with error: cl : Command line error D8021 : invalid numeric argument '/Wextra'.
> 
> Also makes strip command compatible with Darwin OS

PR #502, opened 2022-10-26
Fix set_seed signature in stdlib.fc

AminRezaei0x443:
As described in TVM Documentation we have:

> **F814 — SETRAND (x – )** sets the random seed to unsigned 256-bit Integer x.

But the `set_seed` signature in `stdlib.fc` is incorrect and may confuse developers:
```
int set_seed() impure asm "SETRAND";
```
This PR simply changes the signature to the proper one.

Note: As I wasn't sure whether it's better to open PR into `master` or `testnet`, I thought `master` would be better.


EmelyanenkoK:
Thanks! It is already fixed in `testnet`: https://github.com/ton-blockchain/ton/commit/3d9a16558679ca48ef5616c3518a3c6fd72a0220, will be merged to `master` soon.

PR #507, opened 2022-10-27
Remove pow-miner

zloylos:
Givers mining is over, so it's a good time to remove the legacy code to not confuse users.


andreypfau:
some test networks still use manners to issue test coins, removing this is not good

PR #536, opened 2022-11-29
Fix finding blocks by lt and ut in archive manager

SpyCheese:


awesome-doge:
I feel that my archive node can do this test.

PR #558, opened 2022-12-16
Fix complex funC setglob cases

krigga:
Currently, funC fails to compile any kind of assignment expression involving global variable assignment that isn't of the form `global_var = rvalue;`. It fails to compile modifying methods on global variables and/or assignments to tensors containing global variables, for example:

```
global cell ctx_dict;

(cell, (slice, int)) ~udict_delete_get?(cell dict, int key_len, int index) asm(index dict key_len) "DICTUDELGET" "NULLSWAPIFNOT";

() recv_internal() {
  (slice v, int ok) = ctx_dict~udict_delete_get?(1, 0);
  ~dump(v);
  ~dump(ok);
}
```

Compilation of this file fails with an obscure error:
```
lvalue expression constructor is 9
fatal: cannot compile lvalue expression with unknown constructor
```

This PR aims to fix this.

### DIFF
--- a/crypto/func/func.h
+++ b/crypto/func/func.h
@@ -922,7 +922,7 @@ struct Expr {
   }
   int define_new_vars(CodeBlob& code);
   int predefine_vars();
-  std::vector<var_idx_t> pre_compile(CodeBlob& code, bool lval = false) const;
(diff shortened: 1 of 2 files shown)


EmelyanenkoK:
Closed in favor of #574 which contains this fix as well.

PR #573, opened 2022-12-27
Disable liteServer.getState method

SpyCheese:


thedemons:
Can anyone explain why this was disabled?

PR #594, opened 2023-01-21
`ErrorCode::notready` for "block is not applied":

andreypfau:


EmelyanenkoK:
Scheduled to be merged in February.

PR #595, opened 2023-01-23
Fix error message for unpacking block header

andreypfau:


EmelyanenkoK:
Scheduled to be merged in February.

PR #611, opened 2023-02-04
Invalid cell declaration fix

andreypfau:
All serializable cells MUST be declared with prefix `^` like this: `foo:^Cell`


andreypfau:
All serializable cells MUST be declared with prefix `^` like this: `foo:^Cell`

PR #618, opened 2023-02-15
Make funcfiftlib compilation compatible with modern compilers

dvlkv:


crypto/funcfiftlib/funcfiftlib.cpp
```diff
@@ -99,6 +100,40 @@ td::Result<std::string> compile_internal(char *config_json) {
...
+/// @param _data The data for the callback (a string).
+/// @param o_contents A pointer to the contents of the file, if found. Allocated via malloc().
+/// @param o_error A pointer to an error message, if there is one. Allocated via malloc().
+///
+/// The callback implementor must use malloc() to allocate storage for
+/// contents or error. The callback implementor must use free() to free
+/// said storage after func_compile returns.
+///
+/// If the callback is not supported, *o_contents and *o_error must be set to NULL.
+typedef void (*CStyleReadFileCallback)(char const* _kind, char const* _data, char** o_contents, char** o_error);
```
SpyCheese, line 115: Consider adding `void *context` parameter to `CStyleReadFileCallback` and `func_compile`. It is a traditional way of passing callbacks to functions in C (e.g. [qsort_s](https://en.cppreference.com/w/c/algorithm/qsort)).
However, this is unneeded if it's known for sure that users of this library will only use simple callbacks (without additional state).

EmelyanenkoK: Since this change requires some updates of js code already used in many projects, it is decided to postpone [fix](https://github.com/dvlkv/ton-blockchain/commit/388036b10f619e729cdc16d3c8209fc5fb8f3258) till next update of js-wrapper

PR #621, opened 2023-02-23
Add cross-platform Linux and macOS binaries (statically compiled with NixPkgs) + wasm artifacts

neodiX42:
* new x86-64 Linux binaries are fully cross-platform across major x86-64 Linux distributions.
* arm64 Linux binaries can be assembled the same way, but workflow fails due to GH time limit per job (will be solved later, see `ton-aarch64-linux.yml`);
* x86-64 macOS binaries also cross-platform across x86-64 macOS versions (tested with macOS 11.7 and above)
* arm64 / M1 macOS binaries can also be assembled using NixPkgs, but GH does not support macOS with ARM architecture (see `ton-aarch64-macos.yml`)
* added GH action to build TON WASM binaries -` fift.js, func.js, funcfiftlib.js, tlbc.js` and `emulator-emscripten.js`.
* Thanks to @tonthemoon for his contribution to this commit where he initially introduced and compiled binaries with Nix Package Manager.
* remove compilation against Ubuntu 18.04.
* add standalone artifacts to the release, so it would be possible to download them one by one.


Kraviscloud:
#622

Kraviscloud:
#622

Kraviscloud:
#622

Kraviscloud:
> * new x86-64 Linux binaries are fully cross-platform across major x86-64 Linux distributions.
> * arm64 Linux binaries can be assembled the same way, but workflow fails due to GH time limit per job (will be solved later, see `ton-aarch64-linux.yml`);
> * x86-64 macOS binaries also cross-platform across x86-64 macOS versions (tested with macOS 11.7 and above)
> * arm64 / M1 macOS binaries can also be assembled using NixPkgs, but GH does not support macOS with ARM architecture (see `ton-aarch64-macos.yml`)
> * added GH action to build TON WASM binaries -` fift.js, func.js, funcfiftlib.js, tlbc.js` and `emulator-emscripten.js`.
> * Thanks to @tonthemoon for his contribution to this commit where he initially introduced and compiled binaries with Nix Package Manager.
> * remove compilation against Ubuntu 18.04.
> * add standalone artifacts to the release, so it would be possible to download them one by one.

# 622

PR #641, opened 2023-03-07
Add namespaces to Fift

EmelyanenkoK:


crypto/fift/HashMap.cpp
```diff
@@ -0,0 +1,368 @@
...
+  switch (tp_) {
+    case Type::t_int:
+      return value<td::CntInt256>();
+    case Type::t_atom:
+      return value<vm::Atom>();
+    case Type::t_string:
+    case Type::t_bytes:
+      return {value<td::Cnt<std::string>>(), tp_ == Type::t_bytes};
+    default:
+      return {};
```
SpyCheese, line 57: Keys of unsupported types are converted to null. `hmap!+` with an unsupported key will silently do nothing. It's better to throw an exception in this case.

crypto/fift/HashMap.h
```diff
@@ -0,0 +1,306 @@
...
+  Ref<Hashmap> replace_internal(const DictKey& key, const vm::StackEntry& value, bool& found) const;
+  static void insert(Ref<Hashmap>& root, const DictKey& key, vm::StackEntry value, long long y);
+  static long long new_y();
+};
+
+struct HashmapIdx {
+  Ref<Hashmap>& root_;
+  DictKey idx_;
+  template <typename... Args>
+  HashmapIdx(Ref<Hashmap> root, Args&&... args) : root_(root), idx_(std::forward<Args>(args)...) {
```
SpyCheese: Pass `Ref<Hashmap> &root` instead of `Ref<Hashmap> root`.
If `root` is passed by value, `root_` will refer to parameter `root` which is destroyed after the constructor.

PR #648, opened 2023-03-13
Add modifying version for `skip_dict`

behrang:
I tried to skip extra currencies dictionary in the message header like this:

```
cs~skip_dict();
```

but found out that the modifying version is not in the stdlib.

With current stdlib I have to change the above code to this:

```
cs = cs.skip_dict();
```

which doesn't look good alongside other lines working on the slice.


behrang:
I retract this, since in my latest code I haven't used it at all. Also `load_dict` can be used instead of this if needed and this way we can have a slimmer stdlib.

PR #650, opened 2023-03-14
Fix `compute_data_size` functions in stdlib

behrang:
Changed order of functions to match the documentation. Also replaced `cell` with `slice` in `slice_compute_data_size?` to fix it.


behrang:
This is a breaking change, but the first parameter of the old `slice_compute_data_size?` function was wrong and I think there is no problem in changing it to the fixed version.

PR #668, opened 2023-04-11
add word result for run shell

ilyar:


EmelyanenkoK:
It is unlikely we will allow to run shell commands from fift:
1. fift is good as DSL, but not that handy for general purposes
2. since it's syntax/logic is alien to most developers, it is relatively easy to conceal exploit in the script

So, for complex scenarios and scripts which require interaction with OS/network/other scripts, it is more expedient to use python/js/rust with TON-specific libraries.

PR #669, opened 2023-04-11
Enable more verbose TVM stack dump

dungeon-master-666:
- Print Continuation constructor on stack dump, instead of `???`
- Add flag `DumpStackVerbose` which enables printing serialized boc for Cell and Continuation


ex3ndr:
Is it possible to also move vm::VmLog::DumpStack to the next verbosity level? in my usecase (coverage monitoring) they baloon logs for no reason

PR #670, opened 2023-04-14
add missing VM opcode definition

ilyar:
https://github.com/ton-blockchain/ton/blob/e6f2205dbf5c1b8e4e7320feebadb979838327f1/crypto/vm/tonops.cpp#L725-L728


EmelyanenkoK:
It can not be done this way, since TVM update is consensus critical (network will be halted if more than 1/3 but less than 2/3 of nodes will update and some contract will use new opcodes).
We will think about addition of these opcodes to upcoming [tvm update](https://github.com/ton-blockchain/TEPs/pull/88).

EmelyanenkoK:
Sorry, I misread this issue and decided it suggests to add new opcodes to TVM.
We will consider addition new asm-fif declarations.

PR #710, opened 2023-05-27
Add `?` to `dict_delete_get_min?` and `dict_delete_get_max?`

behrang:
I checked all functions in stdlib, and all of the functions that returns a boolean have a ? suffix, except for dict_delete_get_min and dict_delete_get_max functions, which together with udict and idict versions and modifying versions are a total of 12 functions.

This PR adds the missing `?`. It should not be a breaking change, since old users are using a copy of this stdlib and will not be affected.


behrang:
Sorry, created it on the mainnet, will close and create another one on the testnet branch.

PR #717, opened 2023-05-31
add RustGenerator

ilyar:


andreypfau:
I think it's better to add also to https://github.com/tdlib/td/tree/master/td/generate

MrG9090 (approved):

PR #725, opened 2023-06-08
Fix windows build & remove arm64 ubuntu gh action

neodiX42:
sync testnet with master for windows build; remove long lasting arm64 gh action


Toezaw:
neodiX42:testnet

PR #746, opened 2023-07-15
Fix `advance_ext`

tvorogme:
Such TLB:

```
_ a:(## 64) b:^# = A;
_ a:A = T;
```

Will generate:

<hr/>

```
bool skip(vm::CellSlice& cs) const override {
  return cs.advance_ext(0x10040);
}
```

Which has arguments: `bits: bits_refs >> 16` and `refs: bits_refs & 0xffff`

```
bool CellSlice::advance_ext(unsigned bits_refs) {
  return advance_ext(bits_refs >> 16, bits_refs & 0xffff);
}
```

But it'll also generate:

```
bool T::unpack(vm::CellSlice& cs, T::Record& data) const {
  return cs.fetch_subslice_ext_to(0x10040, data.a);
}
```

Which has arguments: `bits: size & 0xffff` and `refs: size >> 16`

```
td::Ref<CellSlice> CellSlice::fetch_subslice_ext(unsigned size) {
  return fetch_subslice(size & 0xffff, size >> 16);
}
```

I've checked all current `block.tlb` scheme and as I understand bagged `skip` methods are not calling in `cpp` code. But `block-auto.cpp` contains invalid `skip` methods for 59 types. This can lead to random skip size of bits & refs for types.

### DIFF
--- a/crypto/vm/cells/CellSlice.cpp
+++ b/crypto/vm/cells/CellSlice.cpp
@@ -264,7 +264,7 @@ bool CellSlice::advance_ext(unsigned bits, unsigned refs) {
 }
 
 bool CellSlice::advance_ext(unsigned bits_refs) {
-  return advance_ext(bits_refs >> 16, bits_refs & 0xffff);
+  return advance_ext(bits_refs & 0xffff, bits_refs >> 16);
 }
 
 // (PRIVATE)


05373160476:
tvorogme:patch-5

KaurkerDevourer:
Probably got something close:
```
#7  0x00007ffff7c8c4b1 in std::terminate() () from ./libton_crypto_lib.so
#8  0x00007ffff7c8d033 in __cxa_throw () from ./libton_crypto_lib.so
#9  0x00007ffff7981713 in vm::load_cell_slice_impl(td::Ref<vm::Cell>, bool*) () from ./libton_crypto_lib.so
#10 0x00007ffff7981a00 in vm::load_cell_slice(td::Ref<vm::Cell> const&) () from ./libton_crypto_lib.so
#11 0x00007ffff78a14e4 in block::gen::BlockInfo::cell_unpack(td::Ref<vm::Cell>, block::gen::BlockInfo::Record&) const () from ./libton_crypto_lib.so
#12 0x0000555555558c7e in compress (data=...) at solution.cpp:37
#13 0x000055555555940d in solve (in=...) at solution.cpp:75
#14 0x0000555555559627 in main (argc=2, argv=0x7fffffffdc98) at solution.cpp:86
```
I am quite sure, that std::terminate should not be called in any kind of parsing.
Got this with following code:
```
block::gen::BlockInfo::Record state_update_rec;
block::gen::BlockInfo().cell_unpack(block_rec.state_update, state_update_rec);
```

SpyCheese (approved):

PR #751, opened 2023-07-21
Fix merkle_update tag

tvorogme:
Currently, `MERKLE_UPDATE`, stores `#04`
https://github.com/ton-blockchain/ton/blob/testnet/crypto/vm/cells/CellBuilder.cpp#L144


tvorogme:
BTW it may be useful to add tag check [here](https://github.com/ton-blockchain/ton/blob/master/validator/impl/shard.cpp#L193)

tvorogme:
https://github.com/ton-blockchain/ton/pull/1479

PR #760, opened 2023-07-28
Fix error: control reaches end of non-void function

vitlav:


DanShaders:
Fixed in #2229 in 22aa313

PR #785, opened 2023-10-24
Support wallet-v4 in tonlib

SpyCheese:


LeonidBrovkin:
It's time to do it. Yesterday I made an analog of /tutorials/01-wallet from ton-community.github.io with using of tonlib and got the same diff.

By the way, maybe we need to add tl-parser from github.com/vysheng/tl-parser.git as a submodule and to remove generated tlo from the source tree?

PR #786, opened 2023-10-25
CI: fix macOS and tonlib Android GitHub actions in testnet

neodiX42:
GitHub action that builds WASM libraries fails randomly with different GitHub image runner versions. 
This will be fixed with a separate PR.


Slowalki:
.github/script/fift-func-wasm-build-ubuntu.sh

PR #853, opened 2024-01-10
add liteserver method getLibraryExt

yungwine:
### Why do we need this method?

1. Currently there is no way to get library publishers using Liteserver. `getLibraryExt` allows you to ask Liteserver for publishers using `mode` == 1 (otherwise if there is a big hashmap it will be pruned).
2. When you use getLibraries method and liteserver returns you a library code, you can't know is it actually in the blockchain or not. `getLibraryExt` also gives you a proof that you can check and validate that the library is actually in the masterchain state.
3. User will be able to get library state for specified block. This feature for example allows you to get exact block when library has been deployed (e.g. using binary search with archive ls).


EmelyanenkoK:
Merged in https://github.com/ton-blockchain/ton/pull/934

PR #895, opened 2024-02-09
Fix two steps cross-compilation

neodix42:
This fix allows cross-compile of TON on x86-64 Linux for aarch64, using two steps compilation (where first step is generation of auto content and the second - nix cross-compilation)


neodix42:
postponed to better times, when nix.org fixes some bugs related to cross-compile for other platforms.

PR #897, opened 2024-02-10
Add my_code function to stdlib

behrang:


EmelyanenkoK:
We are currently working on stdlib.fc update (as well as some updated of func), that already include this change:
https://github.com/ton-blockchain/ton/pull/874

behrang:
That's great. I wasn't aware of it. I'll close mine.

Omidkalhor65 (approved):

PR #918, opened 2024-02-25
Remove `pragma once` from cpp

hey-researcher:


stenolog:
@hey-researcher , you should at least state a rationale for doing this.

hey-researcher:
> @hey-researcher , you should at least state a rationale for doing this.

We should remove `#pragma once` from a .cpp file because it is unnecessary and misleading; this directive is intended for header files (.h or .hpp) to prevent multiple inclusions, not for source files (.cpp) which are not typically included in other files.

PR #957, opened 2024-04-08
Fix setting due_payment in storage phase

EmelyanenkoK:


common/global-version.h
```diff
@@ -19,6 +19,6 @@
 namespace ton {
 
 // See doc/GlobalVersions.md
-const int SUPPORTED_VERSION = 6;
```
Jimvilcaaa, line 22: 0

crypto/block/transaction.h
```diff
@@ -82,6 +82,7 @@ struct StoragePhaseConfig {
   td::RefInt256 freeze_due_limit;
   td::RefInt256 delete_due_limit;
   bool enable_due_payment{false};
+  int global_version = 0;
```
Jimvilcaaa, line 85: 5

common/global-version.h
Dymas7548: - [ ]

ahmadrashdy14 (commented):
ahmadrashdy14@gmail.com

PR #979, opened 2024-05-05
Fix lite client crashing when block lookup error

hey-researcher:
Here is a line: https://github.com/ton-blockchain/ton/blob/037053fffec58dedf8e72051fb695a4a42d3d9f3/tonlib/tonlib/TonlibClient.cpp#L5567C60-L5567C77

If the liteserver did not succeed in looking up the block, `.ok()` will perform a check and crash the app.


jamescarter-le:
Any chance of getting this PR merged or someone from TON taking a look at this, as causing crashes apps consuming this library.

hey-researcher:
@EmelyanenkoK could check it please

SpyCheese (approved):

PR #991, opened 2024-05-13
broadcast block candidates in public overlay

fasttinykoala:
This commit will allow validators of masterchain to include the new blocks of shardchain faster, because they will not have to wait to download the newest block. The downside of extra block broadcasts doesn't look too problematic for now.


andreypfau:
This PR contains a lot of cosmetic code-formatting changes, which do not relate to the essence of the proposal itself

PR #1005, opened 2024-05-28
Limit checked external messages per address

EmelyanenkoK:


jilio:
Hey @EmelyanenkoK Can you please explain the meaning of these changes? Does it mean that you cannot send multiple transactions with external messages per address in a short period of time at all? Will the transactions be bounced?

PR #1020, opened 2024-06-11
Fix TON compilation on Windows (github only)

neodix42:
With a new OS image runner update for Windows Server 2022, libsodium library behaves weirdly and compilation cannot proceed. At the same time on other Windows machines, it works fine. A temporary solution is to switch to an older Windows image runner.


awesome-doge (approved):
Test passed

截圖 2024-06-22 下午3 16 42

awesome-doge:
There is nothing wrong with your @neodix42 code

But the server of download.libsodium.org is not stable

There is about a 1/2 chance that the compilation will fail because download.libsodium.org cannot be downloaded.

I think we can add some code that attempts to reinstall libsodium to improve the stability of the code.

截圖 2024-06-22 下午3 51 05

neodix42:
Hi awesomedoge,
No, there is definitely an issue with the latest github windows runner. The problem with download.libsodium.org is just an unpleasant rendezvous with comparison to that.

assembly/native/build-windows-2019.bat
```diff
@@ -0,0 +1,221 @@
...
+  echo Can't install secp256k1
+  exit /b %errorlevel%
+)
+cd ..
+) else (
+echo Using secp256k1...
+)
+
+
+curl  -Lo libsodium-1.0.18-stable-msvc.zip https://download.libsodium.org/libsodium/releases/libsodium-1.0.18-stable-msvc.zip
```
awesome-doge:
```suggestion
curl --retry 5 --retry-delay 10 -Lo libsodium-1.0.18-stable-msvc.zip https://download.libsodium.org/libsodium/releases/libsodium-1.0.18-stable-msvc.zip
```

awesome-doge (changes requested):

EmelyanenkoK (approved):

PR #1285, opened 2024-10-18
Updated upload-artifact action version

awesome-doge:
The version of the upload-artifact action used in the workflow has been updated from v1 to v3. This change is expected to improve performance and reliability.


Rafazzzz:
### __

PR #1298, opened 2024-10-21
Optimize store_slice function to reduce gas consumption

zqy-691001:
Certainly! Here's a detailed explanation of the optimization process for the `store_slice` function in the TON Blockchain's `stdlib.fc` file:

---

**Overview:**

The `store_slice` function within the `stdlib.fc` file of the TON Blockchain has been identified as an area for potential optimization. The current implementation leads to unnecessary gas consumption due to the order of operations in the TON Virtual Machine (TVM). This document outlines the proposed changes, the expected benefits, and the steps to contribute this optimization back to the project.

**Current Issue:**

The existing `store_slice` function is implemented as follows in C:

```func
builder store_slice(builder b, slice s) "STSLICER";
```

This results in the following assembly instructions:

```assembly
execute SWAP
execute STSLICER
```

The gas consumption for this sequence is 44 units (18 for SWAP + 26 for STSLICER).

**Proposed Optimization:**

By rewriting the function directly in assembly, we can achieve the desired operation order without the need for an additional SWAP operation:

```func
builder store_slice(builder b, slice s) asm(s b) "STSLICE";
```

This change results in a more efficient single assembly instruction:

```assembly
execute STSLICE
```

With this optimization, the gas consumption is reduced to just 18 units, which is a significant improvement over the original implementation.

**Benefits of the Optimization:**

1. **Reduced Gas Consumption:** The most immediate benefit is the reduction in gas consumption from 44 units to 18 units. This can lead to cost savings for users interacting with smart contracts on the TON Blockchain, as they will require less gas to execute the `store_slice` operation.

2. **Improved Efficiency:** The optimized function is more efficient in terms of both gas usage and execution speed, as it eliminates the unnecessary SWAP operation.

3. **Simplified Assembly:** The assembly code is simplified, which can make it easier to understand and maintain.

### DIFF
--- a/crypto/smartcont/stdlib.fc
+++ b/crypto/smartcont/stdlib.fc
@@ -439,7 +439,7 @@ builder store_ref(builder b, cell c) asm(c b) "STREF";
 
 
 ;;; Stores `slice` [s] into `builder` [b]
-builder store_slice(builder b, slice s) asm "STSLICER";
+builder store_slice(builder b, slice s) asm(s b) "STSLICE";
 
 ;;; Stores (serializes) an integer [x] in the range `0..2^120 − 1` into `builder` [b].
 ;;; The serialization of [x] consists of a 4-bit unsigned big-endian integer `l`,


hossen71 (commented):
good

hossen71 (commented):
builder store_slice(builder b, slice s) asm(s b) "STSLICE";

crypto/smartcont/stdlib.fc
hossen71, line 1: 3158025a9a3bf652e2a700db8c8075e99a23e4e7

hossen71 (commented):
Nais

PR #1462, opened 2025-01-07
extend generate-random-id utility...

neodix42:
with -f parameter (to read addr list from a file), just like -a parameter that reads addr list from the command line


EmelyanenkoK (approved):

Mohamadhosein369 (commented):
M.z

PR #1463, opened 2025-01-07
celldb: version 2

birydrad:
- thread safe cache
- parallel commit
- multiple optimizations
- support of key-value merge operations
- improved tests and benchmarks
- in-memory version won't read from key value after start - uses vector in-memory table now
- use rocksdb::WALRecoveryMode::kTolerateCorruptedTailRecords - do not silently ignore errors during recovery


agca52 (commented):
https://github.com/jlord/sheetsee.js/commit/UQC70oJqt3BauiZon2P4_D8vaADnpq-wE2DUic-CyzScAKFj

EmelyanenkoK (approved):

PR #1496, opened 2025-01-26
Fix BUILD_SHARED_LIBS issue

dungeon-master-666:
In case `cmake` is called several times `BUILD_SHARED_LIBS` variable sets to `ON` which causes building dependencies as shared libraries (e.g. absl libs), which is undesirable. To fix it BUILD_SHARED_LIBS [should be set to `OFF` in top level ](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html) CMakeLists.txt.


neodix42 (approved):
reviewed +1

EmelyanenkoK (approved):

PR #1516, opened 2025-02-06
Fix get_prev_blocks_info() at LS getConfigParams [does not affect validators]

EmelyanenkoK:


ort112586-lgtm:
If the goal is exact reproduction of the config dict at the given block, what mode bits should never be forcibly added unless the original request already asked for them? In other words: should the code ever do mode |= something; in the handler, or should it respect exactly what the client sent (perhaps after sanitizing invalid combinations)?


Suppose we remove the forced addition entirely:






// Proposed cleaner version – no forced |= 
auto res = block::ConfigInfo::extract_config(mpb.root(), mode);  // use whatever mode came in
if (res.is_error()) {
  // handle error (perhaps fall back or reject if needed)
}
cfg = res.move_as_ok();

PR #1534, opened 2025-02-25
Fix for tonlib-cli 'transfer' command

Stanislav-Povolotsky:
Fix for '**transfer**' command of **tonlib-cli**.

Before fix:
```bash
transferf 0 1 100000000
synchronization: 100%
Change destination address from bounceable to non-bounceable
synchronization: DONE in 8501.3us
Query {transferf 0 1 100000000} FAILED:
        [Error : 0 : Failed to add extra currencies]
``` 

After fix:
```bash
transferf 0 1 100000000
synchronization: 100%
synchronization: DONE in 38.7ms
Change destination address from bounceable to non-bounceable
Transfer sent: ok {
}
```

### DIFF
--- a/tonlib/tonlib/TonlibClient.cpp
+++ b/tonlib/tonlib/TonlibClient.cpp
@@ -268,7 +268,7 @@ td::Result<td::Ref<vm::Cell>> add_extra_currencies(const td::Ref<vm::Cell> &e1, 
   block::CurrencyCollection c1{td::zero_refint(), e1};
   block::CurrencyCollection c2{td::zero_refint(), e2};
   TRY_RESULT_ASSIGN(c1, TRY_VM(td::Result<block::CurrencyCollection>{c1 + c2}));
-  if (c1.is_valid()) {
+  if (!c1.is_valid()) {
     return td::Status::Error("Failed to add extra currencies");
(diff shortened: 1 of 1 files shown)


bentrnr21:
Thanks for fixing the behavior of the transfer command in tonlib-cli. Could you provide more details on the specific scenarios where this issue occurred and how your changes address them? This would help the community better understand the impact of this fix and ensure similar cases are considered in the future.

Stanislav-Povolotsky:
Scenario: Run the **tonlib-cli** command and try to transfer TON from one wallet to another using the 'transfer' command.
Result: The command is broken and does not work 100% of the time (always fails with the error "_Failed to add extra currencies_")
After this fix: the command works correctly.
  
The proposed change modifies the condition to:
```C++
  if (!c1.is_valid()) {
    return td::Status::Error("Failed to add extra currencies");
  }
```
Now, the block of code inside the if statement will execute only if c1 is not valid. This change inverts the logic, meaning that if **c1 is invalid**, it will return an error status with the message "Failed to add extra currencies". I believe that the author of this code originally wanted to make such a check.

EmelyanenkoK (approved):

PR #1588, opened 2025-03-30
Avoid CellBuilder during deserialization in CellSerializationInfo

DanShaders:
Another free 2.5% of speed-up on block verification.


Sakzz1994 (commented):
Leave

EmelyanenkoK (approved):

PR #1603, opened 2025-04-03
Fix failing windows build

neodix42:


CMakeLists.txt
```diff
@@ -77,6 +77,7 @@ endif()
 
 if (WIN32)
   message("Add wingetopt")
+  set(CMAKE_POLICY_VERSION_MINIMUM "3.10")
```
DanShaders: This sets `CMAKE_POLICY_VERSION_MINIMUM` for everything that follows. While I doubt this will break anything, it's somewhat nicer to introduce a new variable scope using an "immediately invoked" function and set the variable there  (as I've done with `crc32_scope`).

DanShaders (commented):
LGTM

EmelyanenkoK (approved):

PR #1638, opened 2025-04-26
Update docker-ubuntu-image.yml

LukeLarge:


.github/workflows/docker-ubuntu-image.yml
```diff
@@ -1,4 +1,199 @@
-name: Docker Ubuntu 22.04 image
```
LukeLarge, line 1: should update

PR #1645, opened 2025-04-29
Tolk v0.12: structures, generics, and methods

tolk-vm:
Finally! Or maybe *not yet*?

This update brings Tolk **one step closer** to its final form by introducing:

# Notable changes in Tolk v0.12

1. Structures – named data representations, replacing raw tensors
2. Generics – reusable, type-safe abstractions
3. Methods — extension functions for any custom or built-in type
4. Lots of long stdlib functions were renamed to short methods
5. Fift output now contains original .tolk lines as comments

While auto-packing to/from cells is coming some day, structures already replace unreadable tensors and enable powerful generic types.

# Structures: `struct A { ... }`

Looks like TypeScript — but works in TVM!
```
struct Point {
    x: int;
    y: int;
}

fun calcMaxCoord(p: Point) {
    return p.x > p.y ? p.x : p.y;
}

// declared like a JS object
var p: Point = { x: 10, y: 20 };
calcMaxCoord(p);

// called like a JS object
calcMaxCoord({ x: 10, y: 20 });

// works with shorthand syntax
fun createPoint(x: int, y: int): Point {
    return { x, y };
}
```

* a struct is just **a named tensor**
* `Point` is identical to `(int, int)` at the TVM level
* field access `p.x` works like accessing tensor elements `t.0`, for reading and writing

### How are structures represented at runtime?

* structs are stored on the stack exactly like tensors
* returning a struct is the same as returning a tensor
* passing a struct as an argument behaves like passing a tensor (yes, all its fields)
* nullable structs work like nullable tensors, adding a nullability flag (unless they contain only one non-nullable field)

This means **no bytecode overhead** — you can replace unreadable tensors with clean, structured types.

<table>
<thead>
<th>Before</th>
<th>After</th>
</thead>
<tr>
<td valign=bottom>

```
fun loadData(): (int, slice, int, coins?) {
    var cs = getContractData().beginParse();
    return (
        cs.loadInt(32),   // counter value
        cs.loadAddress(), // owner address
        cs.loadUint(64),  // updated time
        cs.loadCoins(),   // last cost
    );
}

get lastUpdateTime() {
    var (_, _, updatedTime: int, _) = loadData();
    return updatedTime;
}
```

</td>
<td>

```
struct StoredInfo {
    counterValue: int;
    ownerAddress: slice;
    updatedTime: uint64;
    lastCost: coins?;
}

fun loadData(): StoredInfo {
    var cs = getContractData().beginParse();
    return {
        counterValue: cs.loadInt(32),
        ownerAddress: cs.loadAddress(),
        updatedTime: cs.loadUint(64),
        lastCost: cs.loadCoins(),
    }
}

get lastUpdateTime() {
    return loadData().updatedTime;
}
```

</td>
</tr>
</table>

### Declaration & Initialization

* fields can be separated by ; or , (both are valid, like in TypeScript)
* trailing commas are allowed
* default values are supported

When creating a structure, you can specify `StructName { ... }` or simply `{ ... }` if the type is clear from the context (e.g., return type or assignment):
```
var s: StoredInfo = { counterValue, ... };
var s: (int, StoredInfo) = (0, { counterValue, ... });

// also valid, but not preferred
var s = StoredInfo { counterValue, ... };
```

### Nested objects

Without any suprpise, they behave like nested tensors:
```
struct ExtraPayload {
    f1: int;
    f2: int;
}

struct StoredInfo {
    ...
    more: ExtraPayload;  // like (int, int) on a stack
}

var info: StoredInfo = {
    ...
    more: { f1: 0, f2: 1 }
}

info.more.f1    // both for reading and writing
```

### Structs with one field have no overhead

```
// no overhead, it's just `int` on a stack
struct UserId {
    value: int;
}

var id: UserId = { value: 10 };   // on a stack: 10
```

Structs with one field are extremely useful combined with methods.

### Structs and smart casts

Smart casts work for struct fields exactly as expected:
```
if (info.nullableInt != null) {
    info.nullableInt + 3;  // ok, int + int
}
```

And of course, in case of any modification, compiler notices it and drops unsafe smart casts.

### Default values for fields

They are supported, even for complex types.
```
struct DefDemo {
    f1: int = 0;
    f2: int? = null;
    f3: (int, coins) = (0, ton("0.05"));
}

var d: DefDemo = {};         // ok
var d: DefDemo = { f2: 5 };  // ok
```

You should specify all required fields creating an object. There is one exception, though: if a field has `never` type, it can be missed out of a literal while creation. This strange rule becomes useful combined with generics defaults, read below.

### Structs are nominal types

If struct A and B have the same shape, they are **not assignable to each other**, they are different types — unlike type aliases.

```
struct A { v: int }
struct B { v: int }

aObject = bObject;   // error, can not assign `B` to `A`
```

### Structs and global variables

There is a common pattern in FunC to introduce many global variables for storage. When you load storage, you populate them, and then use them in various functions throughout the code.
```
// FunC
global slice st::wallet;
global int st::version;
global int st::caller_fee;
...

// loadData() doesn't return anything,
// it only populates variables
```

You might ask: okay, I want to rewrite it with structures, I'll create a struct for a state, shall I place an object to global?
```
struct State {
    wallet: slice;
    version: int;
    callerFee: int;
    ...
}

global st: State; // ???
```

The answer is: you can, but **it's not recommended**. The reason of saving storage into globals on load is NOT optimization, BUT because it's very unhandy to return them from a function and to pass them around the contract code. 

With structures, it becomes very handy to return/pass an object. 

```
fun loadStorage(): State {
   ...
}

// you can pass a whole
fun handleSomething(state: State, ...) {
   state.callerFee ...
}

// or exact fields you need in a function 
fun handleSomething(callerFee: int, ...) {
   ...
}
```

### Structs and overhead of passing them

You might ask: but if a struct has 10 fields, and I pass it to a function, then under the hood, 10 fields are copied onto a stack! Whereas with globals, nothing is copied.

While in some cases this logic is true, remember, that your struct is **already** on a stack. 
```
fun loadData(): StoredInfo {
    return {
        counterValue: cs.loadInt(32),
        ...
        // nothing is copied, stack is left as is
        // and no globals modifications!
    }
}

fun saveData(st: StoredInfo);

// somewhere
var data = loadData();   // it's on a stack
data.counterValue;       // use and modify it
saveData(data);          // nothing is copied, it's on a stack already
                         // (just needs to be arranged to satisfy fields order)
```

Tip: don't forget to `@inline` simple functions/methods until the compiler does this automatically. In the future, compiler will automatically inline functions (at AST level, not at Fift as `@inline` does, it will be more effective). So, there definitely will be no overhead of passing a struct as a whole.

### Structs and methods

Methods can be declared for a struct as standalone extension functions:
```
// instance method, accepts self
fun Point.getX(self) {
    return self.x
}

// static method, does not accept self
fun Point.create0(): Point {
    return { x: 0, y: 0 }
}
```

Using this pattern, you can declare methods not only for structures, but for any type (including built-ins and unions). Read below, the "Methods" section.

### Limitations of structures

* no readonly fields
* no inheritance
* since objects are placed on a stack, recurring them `struct A { v: A }` is not allowed

# Generics `struct<T>` and `type<T>`

Tolk now supports generics, resembling a mix of TypeScript and Rust. They exist only at the type level (no runtime cost).

```
struct Container<T> {
    isAllowed: bool;
    element: T?;
}

struct Nothing {}

type Wrapper<T> = Nothing | Container<T>;
```

Example usage:
```
fun checkElement(c: Container<T>) {
    return c.element != null;
}

var c: Container<int32> = { isAllowed: false, element: null };

var v: Wrapper<int> = Nothing {};
var v: Wrapper<int32> = Container { value: 0 };
```

### Type arguments are required

Since it's a generic, you should specify type arguments when using it:
```
fun getItem(c: Container)        // error, specify type arguments
fun getItem(c: Container<int>)   // ok
fun getItem<T>(c: Container<T>)  // ok

var c: Container = { ... }       // error, specify type arguments
var c: Container<int> = { ... }  // ok
```

### Auto-inferring type arguments for an object

When you declare a simple generic function, the compiler can automatically infer type arguments for a call:
```
fun doSmth<T>(value: T) { ... }

doSmth(123);         // T = int
doSmth(cellOrNull);  // T = cell?
```

It also works when a generic function accepts a generic struct:
```
fun doSmth<T>(value: Container<T>) { ... }

doSmth({ item: 123 });         // T = int
doSmth({ item: cellOrNull });  // T = cell?
```

### Generic type aliases

Also possible:

```
type ContainerAlias<T> = Container<T>;
type Container32 = Container<int32>;

type ContainerAlias = Container; // error, specify type arguments
```

### Demo: canonical `Maybe` and `Either`

In practice,
- instead of TL/B `Maybe<T>`, Tolk has `T?`
- instead of TL/B `Either<L,R>`, Tolk has `L | R`

But the type system is now so powerful, that "original" ones can also be implemented. Not necessary in practice, **but it works**!
```
struct Nothing {}
struct Just<T> { value: T }
type Maybe<T> = Nothing | Just<T>;

match (obj) {        // Maybe<cell>
    Nothing => {}
    Just => { obj.beginParse() }
}

struct Left<T> { value: T; }
struct Right<T> { value: T; }
type Either<TLeft, TRight> = Left<TLeft> | Right<TRight>;

match (val obj = getResult()) {   // Either<slice, int8>
    Left => obj.value,      // slice
    Right => obj.value,     // int8
}
```

### Demo: `Response<TResult, TError>`

Even though `Either` isn't needed in practice, a similar structure for handling errors could be useful:
```
struct Ok<TResult> { result: TResult; }
struct Err<TError> { err: TError; }

type Response<R, E> = Ok<R> | Err<E>;

fun tryLoadMore(slice: slice): Response<cell, int32> {
    return ... 
        ? Ok { result: ... } 
        : Err { err: ErrorCodes.NO_MORE_REFS };
}

match (val r = tryLoadMore(inMsg)) {
    Ok => { r.result }
    Err => { r.err }
}
```

### Default type arguments

For any T, you can specify its default values. Although in practice it's not a regular feature, it works:
```
struct Nullable<T = int> {
    item: T? = null;
}

var n = Nullable {};  // T = int, n.item is int?
```

This feature is primarily done to have `never` defaults:
```
struct Strange<F2 = never, F3 = never> {
    f1: int;
    f2: F2;
    f3: F3;
}

fun takeStrange<F2, F3>(v: Strange<F2, F3>) {
    ...
}

takeStrange({ f1: 10, f2: 20, f3: 30 });   // F2=int, F3=int
takeStrange({ f1: 10, f3: 30 });           // F2=never, F3=int
takeStrange({ f1: 10 });                   // F2=never, F3=never
```

Semantically, with `never` defaults, you can create structures with optional fields (`never` occupies 0 slots on a stack). This feature will be used in stdlib for built-in functions later.

# Methods: for any types, including structures

Methods are declared as extension functions, similar to Kotlin. A method can accept the first `self` parameter (then it's an instance method) or not accept it (then it's a static method).

```
fun Point.getX(self) {
    return self.x
}

fun Point.create(x: int, y: int): Point {
    return { x, y }
}
```

Methods can be created **for any type**, including aliases, unions, and built-in types:
```
fun int.isZero(self) {
    return self == 0;
}

type MyMessage = CounterIncrement | ...;

fun MyMessage.parse(self) { ... }
// this is identical to
// fun (CounterIncrement | ...).parse(self)
```

### Methods and `asm` function

They perfectly work, since `self` is just a regular variable:
```
@pure
fun tuple.size(self): int
    asm "TLEN";
```

### `self` is immutable by default

By default, **`self` is immutable**. It means that you can't modify it or call mutating methods. To make `self` mutable, you should explicitly declare `mutate self`:
```
fun Point.assignX(mutate self, x: int) {
    self.x = x;   // without mutate, an error "modifying immutable object"
}

fun builder.storeInt32(mutate self, v: int32): self {
    return self.storeInt(v, 32);
}
```

### Methods for generic structs

Allowed, and created seamlessly. Note, that no extra `<T>` is required: while parsing the receiver type, compiler treats unknown symbols as generic arguments.

```
struct Container<T> {
    item: T;
}

// compiler treats T (unknown symbol) as a generic parameter
fun Container<T>.getItem(self) {
    return self.item;
}

// and this is a specialization for integer containers
fun Container<int>.getItem(self) {
    ...
}
```

Another example:
```
struct Pair<T1, T2> {
    first: T1;
    second: T2;
}

// both <T1,T2>, <A,B>, etc. work: any unknown symbols
fun Pair<A, B>.create(f: A, s: B): Pair<A, B> {
    return {
        first: f,
        second: s,
    }
}
```

Methods for generic structs can even be generic themselves:
```
// ac
fun Pair<A, B>.createFrom<U, V>(f: U, s: V): Pair<A, B> {
    return {
        first: f as A,
        second: s as B,
    }
}

Pair<int?, int?>.createFrom(1, 2);
```

### Methods for any receiver

Similarly, any unknown symbol (typically, `T`) can be used to make a method accepting anything:
```
// any receiver
fun T.copy(self): T {
    return self;
}

// any nullable receiver
fun T?.isNull(self): bool {
    return self == null;
}
```

### Method resolution in case of conflict

When you call `someObj.method()`, multiple methods with the same name may exist and theoretically be acceptable:
```
fun T.copy(self) { ... }
fun int.copy(self) { ... }

someVar.copy();   // ???
```

So, the compiler performs matching to find as precise method as follows:
1) Search for exact type receiver like `int.copy` (most practical cases finish here)
2) Search for non-generic receivers that are acceptable (like `int32.copy` / `int?.copy`)
3) Search for generic receivers except `T` (like `Container<T>.copy`)
4) Search for `T` receivers (`T.copy`)

```
fun int.copy(self) { ... }
fun T.copy(self) { ... }

6.copy()              // int.copy (rule 1)
(6 as int32).copy()   // int.copy (rule 2)
(6 as int32?).copy()  // T.copy with T=int? (rule 4)
```

```
type MyMessage = CounterIncrement | CounterReset;
fun MyMessage.check() { ... }
fun CounterIncrement.check() { ... }

MyMessage{...}.check()         // first (rule 1)
CounterIncrement{...}.check()  // second (rule 1)
CounterReset{...}.check()      // first (rule 2)
```

In case of ambiguity, an error is printed:
```
fun int?.doSmth(self) { ... }
fun int64.doSmth(self) { ... }

var v: int32;
v.doSmth();   // error: no exact match, but two possible acceptable receivers
```

### Function overloads

Currently, there are no function overloads
```
fun f(v: int) {}
fun f(v: slice) {}   // error: duplicate symbol
```

But probably, I'll add overloads in the future, and choosing between them (having arguments types) will be identical to choosing methods (having the receiver type).

### Taking references to generic methods

Having a simple function, you can assign it to a variable. Having a generic function/methods, you can do the same, but you should explicitly specify types.
```
fun genericFn<T>(v: T) { ... }
fun Container<T>.getItem(self) { ... }

var callable1 = genericFn<slice>;
var callable2 = Container<int32>.getItem;
callable2(someContainer32);   // pass it as self
```

# Lots of changes in stdlib: short names and methods

Previously, there were no methods (like in FunC), and symbols `endCell()`, `storeInt32()`, etc. were declared as global-scope functions. 

There was a convention that any global-scope function could be called as a method:
```
// FunC
cell config_param(int x) asm "CONFIGOPTPARAM";

config_param(16);   // ok
16.config_param();  // also ok...
```

So, when we called `b.endCell()`, we actually called a global function `fun endCell(b: builder)`. Since all functions were global-scoped, we had to use long explicit names for them. 
```
someTuple.tupleSize();
// why not someTuple.size()? because it's a global function:
// fun tupleSize(t: tuple): int
```

Now, **Tolk separates functions and methods** like it's done in most languages:
1) functions can NOT be called via dot, only methods can
2) methods can have short names, they don't conflict

<table>
<thead>
<th>Before v0.12</th>
<th>Now</th>
</thead>
<tr>
<td>

```
someCell.cellHash();     // or cellHash(someCell)
someSlice.sliceHash();
```

</td>
<td>

```
someCell.hash();         // the only possible
someSlice.hash();
```

</td>
</tr>
</table>

### Changes in methods

Most functions, intended to be called via dot, are now methods. So, if you called them via dot (most likely), nothing is changed for you. If not, you should update your code.
```
fun builder.storeSlice(mutate self, s: slice): self
    asm "STSLICER";

fun builder.storeCoins(mutate self, x: coins): self
    asm "STGRAMS";

...

b.storeCoins(v);           // ok
storeCoins(mutate b, v);   // error now
```

Some functions that were used as methods and had long names, now have short names. Use `t.push` instead of `t.tuplePush`, `t.first` instead of `t.tupleFirst`.
```
fun tuple.push<T>(mutate self, value: T): void
    asm "TPUSH";

@pure
fun tuple.first<T>(self): T
    asm "FIRST";

...

t.push(123);       // ok
t.tuplePush(123);  // error now
```

Global-scope functions that are semantically not methods, they remain functions. So, if you called them via dot (unlikely), you should update your code.
```
fun sendRawMessage(msg: cell, mode: int): void;

sendRawMessage(msg, 0);    // ok
msg.sendRawMessage(0);     // error now
```

### Changes in contract methods

With the introduction of structures and static methods, we can achieve quite a beautiful convention.

<table>
<thead>
<th>Before v0.12</th>
<th>Now</th>
</thead>
<tr>
<td>

```
getMyAddress();
getContractData();
setContractCodePostponed(cell);
getLogicalTime();
```
</td>
<td>

```
contract.getAddress();
contract.getData();
contract.setCodePostponed(cell);
blockchain.logicalTime();
```
</td>
</tr>
</table>

In stdlib, there is an empty struct `contract` with static methods:
```
struct contract {}

fun contract.getData(): cell
    asm "c4 PUSH";

fun contract.setData(c: cell): void
    asm "c4 POP";

...
```

Similarly, we have `blockchain` "namespace" and some others.

It looks pretty, and since IDE suggest only available methods after dot, it's extremely useful.

### Full list of stdlib renamings

<table>
<thead>
<th>Before</th>
<th>Now</th>
</thead>

<tr>
<td>t.tuplePush(v), tuplePush(t, v)</td>
<td>t.push(v)</td>
</tr>
<tr>
<td>t.tupleFirst(), tupleFirst(t)</td>
<td>t.first()</td>
</tr>
<tr>
<td>t.tupleAt(i), tupleAt(t, i)</td>
<td>t.get(i) or just t.0 etc.</td>
</tr>
<tr>
<td>t.tupleSetAt(v, i), tupleSetAt(t, v, i)</td>
<td>t.set(v, i) or just t.0 = v etc.</td>
</tr>
<tr>
<td>t.tupleSize(), tupleSize(t)</td>
<td>t.size()</td>
</tr>
<tr>
<td>t.tupleLast(), tupleLast(t)</td>
<td>t.last()</td>
</tr>
<tr>
<td>getMyAddress()</td>
<td>contract.getAddress()</td>
</tr>
<tr>
<td>getMyOriginalBalance()</td>
<td>contract.getOriginalBalance()</td>
</tr>
<tr>
<td>getMyOriginalBalanceWithExtraCurrencies()</td>
<td>contract.getOriginalBalanceWithExtraCurrencies()</td>
</tr>
<tr>
<td>getContractData()</td>
<td>contract.getData()</td>
</tr>
<tr>
<td>setContractData(c)</td>
<td>contract.setData(c)</td>
</tr>
<tr>
<td>getContractCode()</td>
<td>contract.getCode()</td>
</tr>
<tr>
<td>setContractCodePostponed(c)</td>
<td>contract.setCodePostponed(c)</td>
</tr>
<tr>
<td>getMyStorageDuePayment()</td>
<td>contract.getStorageDuePayment()</td>
</tr>
<tr>
<td>getMyStoragePaidPayment()</td>
<td>contract.getStoragePaidPayment()</td>
</tr>
<tr>
<td>now()</td>
<td>blockchain.now()</td>
</tr>
<tr>
<td>getLogicalTime()</td>
<td>blockchain.logicalTime()</td>
</tr>
<tr>
<td>getCurrentBlockLogicalTime()</td>
<td>blockchain.currentBlockLogicalTime()</td>
</tr>
<tr>
<td>getBlockchainConfigParam(i)</td>
<td>blockchain.configParam(i)</td>
</tr>
<tr>
<td>random()</td>
<td>random.uint256()</td>
</tr>
<tr>
<td>randomRange(limit)</td>
<td>random.range(limit)</td>
</tr>
<tr>
<td>randomGetSeed()</td>
<td>random.getSeed()</td>
</tr>
<tr>
<td>randomSetSeed(seed)</td>
<td>random.setSeed(seed)</td>
</tr>
<tr>
<td>randomizeBy(x)</td>
<td>random.initializeBy(x)</td>
</tr>
<tr>
<td>randomizeByLogicalTime()</td>
<td>random.initialize()</td>
</tr>
<tr>
<td>c.cellHash(), cellHash(c)</td>
<td>c.hash()</td
</tr>
<tr>
<td>s.sliceHash(), sliceHash(s)</td>
<td>s.hash()</td>
</tr>
<tr>
<td>s.sliceBitsHash(), sliceBitsHash(s)</td>
<td>s.bitsHash()</td>
</tr>
<tr>
<td>c.calculateCellSize(maxCells), calculateCellSize(c, maxCells)</td>
<td>c.calculateSize(maxCells)</td>
</tr>
<tr>
<td>s.calculateSliceSize(maxCells), calculateSliceSize(s, maxCells)</td>
<td>s.calculateSize(maxCells)</td>
</tr>
<tr>
<td>c.getCellDepth(), getCellDepth(c)</td>
<td>c.depth()</td>
</tr>
<tr>
<td>s.getSliceDepth(), getSliceDepth(s)</td>
<td>s.depth()</td>
</tr>
<tr>
<td>b.getBuilderDepth(), getBuilderDepth(b)</td>
<td>b.depth()</td>
</tr>
<tr>
<td>debugPrint(x), x.debugPrint(x)</td>
<td>debug.print(x)</td>
</tr>
<tr>
<td>debugPrintString(s), s.debugPrintString()</td>
<td>debug.printString(s)</td>
</tr>
<tr>
<td>debugDumpStack()</td>
<td>debug.dumpStack()</td>
</tr>
<tr>
<td>beginParse(c)</td>
<td>c.beginParse()</td>
</tr>
<tr>
<td>endCell(b)</td>
<td>b.endCell()</td>
</tr>
<tr>
<td>s.assertEndOfSlice(), assertEndOfSlice(s)</td>
<td>s.assertEnd()</td>
</tr>
<tr>
<td>s.getRemainingRefsCount(), getRemainingRefsCount(s)</td>
<td>s.remainingRefsCount()</td>
</tr>
<tr>
<td>s.getRemainingBitsCount(), getRemainingBitsCount(s)</td>
<td>s.remainingBitsCount()</td>
</tr>
<tr>
<td>s.getRemainingBitsAndRefsCount(), getRemainingBitsAndRefsCount(s)</td>
<td>s.remainingBitsAndRefsCount()</td>
</tr>
<tr>
<td>s.isEndOfSlice(), isEndOfSlice(s)</td>
<td>s.isEnd()</td>
</tr>
<tr>
<td>s.isEndOfSliceBits(), isEndOfSliceBits(s)</td>
<td>s.isEndOfBits()</td>
</tr>
<tr>
<td>s.isEndOfSliceRefs(), isEndOfSliceRefs(s)</td>
<td>s.isEndOfRefs()</td>
</tr>
<tr>
<td>isSliceBitsEqual(a, b), a.isSliceBitsEqual(b)</td>
<td>a.bitsEqual(b)</td>
</tr>
<tr>
<td>b.getBuilderRefsCount(), getBuilderRefsCount(b)</td>
<td>b.refsCount()</td>
</tr>
<tr>
<td>b.getBuilderBitsCount(), getBuilderBitsCount(b)</td>
<td>b.bitsCount()</td>
</tr>
</table>

Low-level functions for TVM dictionaries were not renamed, but all of them became methods of `dict` (it's `cell?` actually). Some day, there will be a type `map<K,V>` with user-frieldly API, and methods like `dict.iDictGetPrevOrEqual` will still exist in `@stdlib/tvm-dicts`.

# No "compute-asm-ltr" when it's safe

In FunC, when an `asm` function had arg_order (swapping stack arguments), passed arguments were calculated not left-to-right, but as `asm` directive suggests:
```
// FunC
() some_asm_function(int a, int b) pure asm(b a) "...";

some_asm_function(f1(), f2());   // f2 was called before f1
```

Since this behavior is unexpected (reverted function called, assignments, etc.), `#pragma compute-asm-ltr` was introduced that always evaluated arguments in the left-to-right order.

When Tolk was forked, since its initial version v0.6, this pragma was always turned on, arguments were evaluated in a predictable ltr order. However, for arguments without side effects (like simple constants), this is less efficient, since stack should be reordered later. That's why code converted to Tolk from FunC line-by-line could consume +0.5-1% more gas than in FunC without this pragma.

Current release contains an optimization to detect whether it's safe to put arguments not in ltr order in advance to prevent extra stack manipulations. For example, `f(1,2,3)` / `b.storeUint(2,32)` is safe, but `f(x,x+=5,x)` / `f(impureF1(), global_var)` is not. 

In real-world contracts, in most cases it's actually safe, so this extra bit of gas consumption over FunC no longer exists.

# Fift output now contains original .tolk lines as comments

Since this update it becomes much easier to dig into Fift assembler produced by the compiler:
* stack comments are aligned
* struct fields are distinct in stack comments
* original lines from .tolk files are present as comments

Example tolk input function:
image

<details>
  <summary>Expand Tolk input</summary>

```
fun demo(initialX: int) {
    var p: Point = { x: initialX, y: 20 };
    var maxCoord = max(p.x, p.y);
    assert(maxCoord > 0, 0xFFFF);
    if (maxCoord > 100) {
        logBigPoint(p);
    }
    return p.x == maxCoord;
}
```
  
</details>

Its Fift code:
image

<details>
  <summary>Expand Fift output</summary>

```
// dev.tolk:8
demo PROC:<{              	//  initialX
  // 9: var p: Point = { x: initialX, y: 20 }
  20 PUSHINT              	//  p.x p.y=20
  2DUP                    	//  p.x p.y=20 p.x p.y=20
  // 10: var maxCoord = max(p.x, p.y)
  MAX                     	//  p.x p.y=20 maxCoord
  // 11: assert(maxCoord > 0, 0xFFFF)
  16 PUSHPOW2DEC          	//  p.x p.y=20 maxCoord '6=65535
  OVER                    	//  p.x p.y=20 maxCoord '6=65535 maxCoord
  0 GTINT                 	//  p.x p.y=20 maxCoord '6=65535 '8
  THROWANYIFNOT           	//  p.x p.y=20 maxCoord
  // 12: if (maxCoord > 100)
  DUP                     	//  p.x p.y=20 maxCoord maxCoord
  100 GTINT               	//  p.x p.y=20 maxCoord '11
  IF:<{                   	//  p.x p.y=20 maxCoord
    s2 s1 PUXC            	//  p.x maxCoord p.x p.y=20
    // 13: logBigPoint(p)
    logBigPoint CALLDICT  	//  p.x maxCoord
  }>ELSE<{
    NIP                   	//  p.x maxCoord
  }>
  // 15: return p.x == maxCoord
  EQUAL                   	//  '12
}>
```
</details>

Actually, the optimizer can join/rearrange some instructions, that's why some original lines may be skipped or even swapped. It's expected, because it's exactly the purpose of a peephole optimizer.

The compiler has options to disable/lines output, both from a command-line and a JS wrapper.

# Related pull requests

* https://github.com/ton-blockchain/convert-func-to-tolk/pull/6
* https://github.com/ton-community/ton-docs/pull/1153
* https://github.com/ton-blockchain/tolk-js/pull/8
* https://github.com/ton-blockchain/tolk-vscode/pull/9


09134643851 (commented):
عنوان رسید: شرکت ***

EmelyanenkoK (approved):

PR #1648, opened 2025-05-04
Remove concept of virtualization level

DanShaders:
But for a single place in test-db, we used virtualization level 0 for non-virtualized cells and level 1 for virtualized ones. With this patch, we do not store this level anymore and instead infer it based on "effective" level (previously known as VirtualizationParameters::level) and the actual level of the cell.

This is mainly done to reduce cognitive complexity of the code. However, as a nice side-effect, changes in VirtualCell give a negligible performance improvement on the verification benchmark.


DanShaders:
TODO:
- [x] Run network with this.
- [ ] ~~Check if there is any speed up on benchmark~~ I don't have benchmark anymore

PR #1655, opened 2025-05-08
Set config in tvm_emulator_emulate_run_method and add separate method to read VM logs

yma-het:


emulator/emulator-extern.cpp
```diff
@@ -672,12 +678,43 @@ const char *tvm_emulator_emulate_run_method(uint32_t len, const char *params_boc
   if (!libs.is_empty()) {
     emulator->set_libraries(std::move(libs));
   }
+
+  if (c7->depth() > 0) {
```
yma-het, line 682: I could linearize this logic with goto, but I have not found usage of this operator in code, and considering controversial nature of it, I've decided to not use it.

EmelyanenkoK:
Closed in favor of https://github.com/ton-blockchain/ton/pull/1717 (merged PR hugely based on this one)

PR #1665, opened 2025-05-11
add missing tag in BlockExtra description

Sild:


DanShaders:
Automatic CRC32 tags is a feature of TL-B. If you parser doesn't support them, use a different parser.

PR #1741, opened 2025-07-07
Tolk v1.0: lazy loading, AST inlining, peephole optimizations, TVM 11

tolk-vm:
We've made it. Tolk is now officially 1.0 — stable, production-ready, and battle-tested.

This release brings together all the work done over the past months: powerful type system, expressive syntax, full auto-serialization, idiomatic message sending — and now, with v1.0, we complete the picture with several on-top features:

1. The magic `lazy` keyword — lazy loading, partial loading, partial updating
2. Auto-detect and inline functions at the compiler level
3. Various optimizations for gas efficiency
4. `onInternalMessage` and `onBouncedMessage`, TVM 11 support
5. Custom pack/unpack serializers for custom types

**And benchmarks!** We complete the picture with "Tolk vs FunC" benchmarks on real contracts.

# Lazy loading, partial loading, partial updating

One magic `lazy` keyword to rule them all.

### Lazy loading, partial loading

Say, you have a `Storage` in a wallet:
```
struct Storage {
    isSignatureAllowed: bool
    seqno: uint32
    subwalletId: uint32
    publicKey: uint256
    extensions: dict
}

fun Storage.load() {
    return Storage.fromCell(contract.getData())
}
```

What does `Storage.load()` do? It unpacks a cell and populates all fields of a struct, checks consistency, etc.

**The magic `lazy Storage.load()` does NOT load a full cell**. Instead, the compiler tracks — what fields you are using, and automatically "picks" requested fields, skipping others.

```
get fun getPublicKey() {
    val st = lazy Storage.load();
    // <-- here "skip 65 bits, preload uint256" is inserted
    return st.publicKey
}
```

That's it! With a single `lazy` keyword you defer loading until being accessed. The compiler tracks all control flow paths and automatically inserts loading points, groups unused fields for skipping, etc. Of course, it works with any types, any combination of fields used in any places within your code — the compiler tracks everything.

### Even deeper than you might think

Say, you have an NFT collection:
```
struct NftCollectionStorage {
    adminAddress: address
    nextItemIndex: uint64
    content: Cell<CollectionContent>
    ...
}

struct CollectionContent {
    metadata: cell
    minIndex: int32
    commonKey: uint256
}
```

And somewhere, you need just to get `content` from a storage (and to get `commonKey` from there). 
```
val storage = lazy NftCollectionStorage.load();
// <-- here just "preload ref" is inserted
val contentCell = storage.content;
```

The first trick: no "skip address, skip uint64" is needed. To get a ref, no need to skip preceding data fields — the compiler knows about this.

Second: okay, we have `contentCell`. How to get `commonKey` from there? The answer: it's a cell => you need to load it... lazily!
```
val storage = lazy NftCollectionStorage.load();

// <-- "preload ref" inserted — to get `content`
// Cell<T>.load() unpacks a cell and gives you T
val content = lazy storage.content.load();

// <-- "skip 32 bits, preload uint256" - to get commonKey
return content.commonKey;
```

A small reminder about `Cell<T>`. These are typed cells, widely used to describe nested references. When you have `p: Cell<Point>`, you can't access `p.x` — you need to load that cell, either with `Point.fromCell(p)` — or (better) with `p.load()`. Both variants can be called with `lazy`.

### Lazy matching

Similarly, when reading a union (an incoming message, for example), you do it with `lazy`:
```
struct (0x12345678) CounterReset { ... }
...
type MyMessage = CounterReset | CounterIncrement | ...

val msg = lazy MyMessage.fromSlice(msgBody);    
match (msg) {
    CounterReset => {
        assert(senderAddress == storage.owner) throw 403;
        // <-- here "load msg.initial" is inserted
        storage.counter = msg.initial;
```

With `lazy` for a union,
1) no union is allocated on a stack, actually: matching and loading is deferred
2) `match` naturally works by a slice prefix (opcode)
3) inside every branch, the compiler inserts loading points, skips unused fields, etc. — all that it does for a struct

That's why **lazy matching is very efficient**. More efficient than `if (op == OP_RESET)` FunC pattern. Described in terms of the type system, it perfectly fits TVM nature without extra stack manipulations.

### Lazy matching and ELSE

Since lazy `match` for a union is done by a prefix (opcode), when none of prefixes match, you can handle it with `else` branch. 

In FunC contracts, there was a common pattern to "ignore empty messages":
```
// FunC-style
if (msgBody.isEmpty()) {
    return;    // ignore empty messages
}
val op = msgBody.loadUint(32);  // because this would throw excno 9

if (op == OP_RESET) { 
    ...
    return; 
}

throw 0xFFFF;  // "invalid opcode"
```

The only reason for handling empty in advance was not to throw "cell underflow" in "loadUint".

With lazy `match`, no need to waste gas in advance. Do all checks in `else`:
```
val msg = lazy MyMessage.fromSlice(msgBody);    
match (msg) {
    CounterReset => { ... }
    ...  // handle all types of a union

    // else - when nothing matched;
    // even corrupted input (less than 32 bits), no "underflow" fired
    else => {
        // ignore empty messages, "wrong opcode" for others
        assert (msgBody.isEmpty()) throw 0xFFFF
    }
}
```

Without manual `else`, unpacking will throw 63 by default (controlled by an extra option `throwIfOpcodeDoesNotMatch` for fromCell/fromSlice). With `else`, you can override any behavior.

Note that `else` in `match` by type is allowed only with `lazy`, because it works on prefixes. Without lazy, it's just a regular union, matched by a union tag (typeid) on a stack.

### Partial updating

It's not the end of magic. `lazy` not only works on reading, but perfectly **works on writing back**.

Say, we load storage, use its fields for assertions, modify one field, and save it back:
```
val storage = lazy Storage.load();

assert (storage.validUntil > blockchain.now()) throw 123;
assert (storage.seqno == msg.seqno) throw 456;
...

storage.seqno += 1;
contract.setData(storage.toCell());   // <-- magic
```

The compiler is smart, and for toCell() it **does not save all fields of a storage**, because we modified only `seqno`. Instead, on loading, after loading `seqno`, it saved "immutable tail," and reused it writing back:
```
val storage = lazy Storage.load();
// actually, what was done:
// - load isSignatureAllowed, seqno
// - save immutable tail
// - load validUntil, etc.

storage.seqno += 1;
storage.toCell();
// actually, what was done:
// - store isSignatureAllowed, seqno
// - store immutable tail
```

No more manual optimizations with intermediate slices — the compiler does all the stuff for you. It can even group unmodified fields in the middle, load them as a slice, and preserve this slice when writing back.

Of course, it works only when a compiler statically resolves behavior within a local scope of a function. If you use globals, split logic into non-inlined functions, this optimization will obviously break, and `lazy` will fall back to regular loading.

### Q: What are disadvantages of `lazy`?

In terms of gas usage, lazy fromSlice is always less or equan than regular fromSlice, because in the worst case, when you access all fields, they are obviously all loaded one by one.

But there is another difference, not related to gas usage. 
* When you do `T.fromSlice(s)`, it unpacks all fields of T and then inserts `s.assertEnd()` (it can be disabled by an option). So, if a slice is corrupted or has extra data, fromSlice will throw.
* `lazy`, of course, "picks" only requested fields and easily works with partially invalid input. For example, `struct Point { x: int8, y: int8 }`, and you use `lazy Point` and `p.x` only. Then an input "FF" (8 bits) is okay, even though `y` is not presented at all. Similarly, "FFFF0000" (16 bits of extra data) is also okay, `lazy` just ignores everything not requested.

Typically, it's not a problem. For storage, you have a guarantee that it's shaped correctly — since your contract manages it. For incoming messages, you typically use all fields (otherwise, why do you need them in structs?). If there is some extra data in input — who cares? The message can be deserialized well, I don't find any error here.

My own opinion is that in the future we'll have a "development TVM mode." Tolk compiler will be able to emit some "validation instructions for dev purposes" that will not affect bytecode and gas usage, but be a part of some "debug symbols," handled by TVM during execution. While developing, you'll see rich errors when something goes wrong (overflow on assignment, invalid tuple indices, etc.). Having such tooling, you'll have much more info when something goes wrong — and it's not about lazy or corrupted data, it's about execution in general.

Probably, that day `lazy` will become a default. For now, it's a separate keyword emphasizing "lazy loading" and a killer feature of Tolk.

# Auto-detect and inline functions at compiler level

Since now, Tolk can inline functions at the compiler level, NOT via `PROCINLINE` by Fift.

```
fun Point.create(x: int, y: int): Point {
    return {x, y}
}

fun Point.getX(self) {
    return self.x
}

fun sum(a: int, b: int) {
    return a + b;
}

fun main() {
    var p = Point.create(10, 20);
    return sum(p.getX(), p.y);
}
```

is compiled just to

```
main PROC:<{
  30 PUSHINT
}>
```

The compiler automatically detects what functions to inline. The attribute `@inline` also forces compiler inlining. The attribute `@noinline` forces a function to stay in a dict, `@inline_ref` remains "inline ref" (perfect for "unlikely" execution paths).

Compiler inlining is efficient in terms of stack manipulations. It works with arguments of any stack width. It works with any functions and methods, except recursive or having `return` in the middle. It works with `mutate` and `self`.

You should not worry that "simple getters" like `fun Point.getX(self) { return self.x }` will require stack reordering. You can extract small functions, they are absolutely zero-cost. You don't think about `@inline` and Fift, because no inlining is deferred to Fift: the compiler will handle everything in advance.

### How does "auto-inline" work?

In two words,
- "simple small functions" are inlined always
- if a function is called only once, it's inlined

Some details. For every function, the compiler calculates "weight" (a heuristic AST-based metric) and usages count. If "weight < THRESHOLD," it's inlined always. If "usages == 1," it's inlined always. Otherwise, there is some empiric formula.

No one prevents you to force `@inline` annotation for "big functions" also. For example, if you know that all usages correspond to hot paths. Sometimes, on the contrary, you'd want to "prevent inlining" with `@inline_ref` for example, even if a function is called once — if it's an unlikely path. Anyway, if you are keen on optimization, cover your contract with gas benchmarks and experiment with inlining, branch reordering, etc.

### What can NOT be auto-inlined?

A function will NOT be inlined (even marked as `@inline`) if:

1) `return` in the middle. If a function has multiple return points, the compiler can't do anything with it, currently. In Tolk v1.x it will be partially resolved for some use cases.

2) A function that appears in a recursive call chain (`f -> g -> f`). Very unlikely in practice.

3) It's used as non-call. For example, you take a reference to it like `val callback = f`

# Various optimizations for gas efficiency

Aside the language itself, Tolk compiler became smarter in low-level stack manipulations. After generating intermediate Fift code, it now performs some extra optimizations guaranteed to be safe.

Some are peephole:
- N LDU + NIP => N PLDU
- SWAP + N STU => N STUR
- SWAP + STSLICE => STSLICER и т.п.
- SWAP + EQUAL => EQUAL (+ other symmetric like MUL, OR, etc.)
- SWAP + LESS/LEQ => GREATER/GEQ 
- NOT + N THROWIFNOT => N THROWIF
- ...

Some are semantics:
- ternary operator to CONDSEL when it's safe
- generate x{...} STSLICECONST instead of PUSHINT + STU (with heuristics, because "always" has a negative effect)
- ...

Also, the compiler can now merge constant storeUint. You no longer need to write `b.storeUint(0x18, 6)`. Now
```
// was inefficient in FunC
b.storeUint(0, 1)  // prefix
 .storeUint(1, 1)  // ihr_disabled
 .storeUint(1, 1)  // bounce
 .storeUint(0, 1)  // bounced
 .storeUint(0, 2)  // addr_none
```
is just translated to
```
b{011000} STSLICECONST
```

It works even with variables, when they turn out to be constant.
```
var x = 0;
var b = beginCell();
b.storeUint(x, 4);
x += 12;
x += x;
b.storeUint(x + 2, 8);
```
is just translated to
```
NEWC
x{01a} STSLICECONST
```

Since it works at IR level, it works even for structures — including their fields:
```
struct Point {
    x: uint32
    y: uint32
}

var p: Point = { x: 10, y: 20 };
return p.toCell();
```
becomes
```
NEWC
x{0000000a00000014} STSLICECONST
ENDC
```
(in the future, Tolk will be able to emit a constant cell here)

This optimization reveals why `createMessage` for unions is so lightweight. The compiler really generates all IF-ELSE and STU, but in a later analysis, they become constant (since types are compile-time known), and everything flattens into simple PUSHINT / STSLICECONST.

Not all optimizations that I want can be done with inside the current codebase: that part is still "inherited" from FunC and is very hard to deal with. Some day, I will rewrite everything, and give myself a space for wider optimization abilities.

# Modern `onInternalMessage` and `onBouncedMessage`

Since Tolk v1.0, you don't need to manually parse `msg_cell` to retrieve `sender_address` or `fwd_fee`. Everything is pretty straightforward:
```
fun onInternalMessage(in: InMessage) {
    in.senderAddress
    in.originalForwardFee
    in.valueCoins   // typically called "msg value"

    in.|   // IDE suggests you
}
```

While an "old-fashioned way" with accepting 4 parameters on a stack still works, the above is preferred. Moreover, it's much more efficient: fields of `InMessage` are mapped onto new TVM-11 instructions.

### A pattern of `onInternalMessage`

Is the following:
1) describe every message in a struct (most likely, with 32-bit opcode)
2) describe a union "allowed messages"
3) `val msg = lazy MyUnion.fromSlice(in.body)`
4) `match (msg)` for every branch, and probably `else`

Do not follow an "old-fashioned way" to extract fwd_fee etc. at the beginning of the function, there is no need of it. Just use `in.smth` on demand.

```
type AllowedMessageToMinter =
    | MintNewJettons
    | BurnNotificationForMinter
    | RequestWalletAddress

fun onInternalMessage(in: InMessage) {
    val msg = lazy AllowedMessageToMinter.fromSlice(in.body);

    match (msg) {
        BurnNotificationForMinter => {
            var storage = lazy MinterStorage.load();
            ...
            storage.save();    
            ...
        }
        RequestWalletAddress => ...
        MintNewJettons => ...
        else => {
            // for example:
            // ignore empty messages, "wrong opcode" for others
            assert (in.body.isEmpty()) throw 0xFFFF
        }
    }
}
```

When it comes to storage, it's an open question: load it once (before match) or lazily in every branch? My own experiments on real contracts state than `lazy` in every branch gives better result (bytecode becomes larger, but gas becomes less). Although in your case, it may be falsy. Feel free to test different strategies, having gas metrics.

```
// a typical storage implementation

struct MinterStorage {
    ...
}

fun MinterStorage.load() {
    return MinterStorage.fromCell(contract.getData())
}

fun MinterStorage.save(self) {
    contract.setData(self.toCell())
}
```

### A separate `onBouncedMessage`

In FunC, you parsed `msg_cell`, read 4-bit flags, and tested for `flags & 1` to check whether a message is bounced.

In Tolk, if you want to handle bounces, you create a separate entrypoint:
```
fun onBouncedMessage(in: InMessageBounced) {
}
```
It's automatically called by the compiler, similarly to this:
```
fun onInternalMessage(in: InMessage) {
    // the compiler inserts this automatically:
    if (MSG_IS_BOUNCED) { onBouncedMessage(...); return; }

    ... // your code
}
```

If you don't declare `onBouncedMessage`, all bounces will be just filtered out:
```
fun onInternalMessage(in: InMessage) {
    // the compiler inserts this automatically:
    if (MSG_IS_BOUNCED) { return; }

    ... // your code
}
```

### Remember about 256 bits on bounces

Currently, TON Blockchain works in such a way that when a message is bounced, only the first 256 bits are present: 0xFFFFFFFF ("bounced prefix") + 224 bits from the originally sent message. That's why you manually control which fields you can read (they fit rest 224 bits), which not.

```
fun onBouncedMessage(in: InMessageBounced) {
    in.bouncedBody.skipBouncedPrefix();   // skips 0xFFFFFFFF
    // handle rest of body, probably with lazy match
}
```

In the near future, hopefully, the TON Core Team will provide functionality to change this behavior and allow you to have full contents of a messages when bounced. Since these plans are vague, they are not reflected in the language. So, for now, the only correct way to manage `in.bouncedBody` is to keep that specifics in mind.

# Custom pack/unpack serializers for custom types

Imagine that you have your own "variable-length string": you encode its len, and then data, like
```
len: (## 8)        // 8 bits of len
data: (bits len)   // 0..255 bits of data
```

To express this, you can create your own type and **define a custom serializer**:
```
type VariadicString = slice

fun VariadicString.packToBuilder(self, mutate b: builder) {
    val nBits = self.remainingBitsCount();
    b.storeUint(nBits, 8);
    b.storeSlice(self);
}

fun VariadicString.unpackFromSlice(mutate s: slice) {
    val nBits = s.loadUint(8);
    return s.loadBits(nBits);
}
```

And just use `VariadicString` as a regular type — everywhere:
```
tokenName: VariadicString
fullDomain: Cell<VariadicString>
```

The compiler inlines your functions every time packing/unpacking takes place.

Method names `packToBuilder` and `unpackFromSlice` are reserved for this purpose, their prototype must be exactly as showed.

# Semicolon in declarations made optional

We made a decision to make global-scope declarations "less noisy" and decided to make a semicolon optional — for type aliases, constants, struct fields, and other declarations:

```
const ERR_UNKNOWN = 123

struct Point {
    x: int8
    y: int8
}

type UserId = int
type MyMessage = A | B

struct Empty
```

Struct fields still can be separated by `,` or `;`, but the desired style is just to use newlines.

Note, that it applies to declarations only, NOT to code blocks. Inside functions, semicolon is still required between statements. 

# "Tolk vs FunC" gas benchmarks

With all the features above, I took several standard contracts from the TON ecosystem (Jetton, NFT, Wallet, etc.) and migrated them **from FunC to Tolk**:
- preserving the original logic and behavior,
- passing the same test suites as the FunC versions,
- but written in idiomatic, expressive Tolk style,
- with **significantly reduced gas costs**.

All of them are presented here: [tolk-bench repository](https://github.com/ton-blockchain/tolk-bench)

As you see, not only the code becomes readable, but gas usage is noticeably reduced. And there are no tricks here. Just clean, consistent logic — whether it's a Jetton or a Wallet. 

The rule is simple: **if you write code the way the language encourages — gas will take care of itself**.

# Related pull requests

* https://github.com/ton-blockchain/tolk-js/pull/11
* https://github.com/ton-community/ton-docs/pull/1313
* https://github.com/ton-blockchain/tolk-vscode/pull/13
* https://github.com/ton-blockchain/convert-func-to-tolk/pull/8

### New repositories

1) **TON Language Server** — https://github.com/ton-blockchain/ton-language-server

A new VS Code extension "TON" supporting TOLK, FunC, Fift, and TL/B. And also — a language server for LSP-based editors. Support for Tolk was implemented by @i582 and covers lots of features: smart code completion, auto imports, go to definition, find usages, renaming, type provider, inlay hints, toolchain detection, and more.

The old VS Code extension "Tolk" was unpublished and won't be maintained.

2) **"Tolk vs FunC" gas benchmarks** — https://github.com/ton-blockchain/tolk-bench

Mentioned above. Demonstrates how real smart contracts look in Tolk — and how efficient they can be.


kambyzsahhamt-lang:
Transaction 0.32 then 0.97 then 0.94. This transaction was done. I played the game for $200 but it didn't pay. The amount of the game is my daughter has cancer and I need to buy medicine. UQDP-paXoqSb3BvRz2cq0V3QqbiXIC5FwAd0ZuKtKGVwKO7S      😭😭😭😭😭😭

kambyzsahhamt-lang:
I don't have to buy medicine, for Christ's sake, I'm tired. The world is so worthless.Ruthless. I begged on Twitter for 25 days, but no one helped me. UQDP-paXoqSb3BvRz2cq0V3QqbiXIC5FwAd0ZuKtKGVwKO7S😭😭😭😭😭😭😭😭😭😭😭

PR #1857, opened 2025-10-27
Bugfix/header checks gen utime msg

Lapo4kaKek:
While using tonlib I hit a header verification error (mode &4). The code compares gen_utime vs utime, but the error messages referenced end_lt/lt; this PR fixes the strings only (no logic change).


Edgar71769:
.

PR #1886, opened 2025-11-13
Tolk v1.2: rich bounced messages, cheap deployment, and a breaking change that you'll love

### LINKED ISSUE #1830: Tolk: `pipe-lazy-load-insertions`: Crash on incorrect argument of `__expect_lazy`
MRE:
```tolk
struct S { x: int8 }
fun main() {
    var s = lazy S.fromSlice("");
    __expect_lazy("[s] load x");
}
```

Output:
```
terminate called after throwing an instance of 'std::out_of_range'
  what():  vector::_M_range_check: __n (which is 2) >= this->size() (which is 2)
zsh: IOT instruction  ./tolk/tolk crash_ast.h_1028_48.tolk
```

Backtrace:
```
   frame #8: 0x000055555560bca1 tolk`tolk::CheckExpectLazyAssertionsVisitor::visit(tolk::Vertex<(tolk::ASTNodeKind)42> const*) [inlined] std::vector<tolk::ASTNodeBase const*, std::allocator<tolk::ASTNodeBase const*>>::_M_range_check(this=size=2, __n=2) const at stl_vector.h:1160:4
    frame #9: 0x000055555560bc64 tolk`tolk::CheckExpectLazyAssertionsVisitor::visit(tolk::Vertex<(tolk::ASTNodeKind)42> const*) [inlined] std::vector<tolk::ASTNodeBase const*, std::allocator<tolk::ASTNodeBase const*>>::at(this=size=2, __n=2) const at stl_vector.h:1201:2
    frame #10: 0x000055555560bc64 tolk`tolk::CheckExpectLazyAssertionsVisitor::visit(tolk::Vertex<(tolk::ASTNodeKind)42> const*) [inlined] tolk::Vertex<(tolk::ASTNodeKind)42>::get_item(this=0x0000555555b396b0, i=2) const at ast.h:1028:48
    frame #11: 0x000055555560bc64 tolk`tolk::CheckExpectLazyAssertionsVisitor::visit(this=0x00007fffffffaa98, v=0x0000555555b396b0) at pipe-lazy-load-insertions.cpp:993:31
    frame #12: 0x00005555555f7e56 tolk`void tolk::visit_ast_of_all_functions<tolk::CheckExpectLazyAssertionsVisitor>() [inlined] tolk::CheckExpectLazyAssertionsVisitor::start_visiting_function(this=0x00007fffffffaa98, fun_ref=<unavailable>, v_function=<unavailable>) at pipe-lazy-load-insertions.cpp:1021:13
```

### LINKED ISSUE #1831: Tolk: `pipe-resolve-identifiers`: Crash on unexpected syntax
MRE:
```
struct A { }
struct B { }
type C = A | B;

@noinline
fun C.onInternalMessage(self) {
    return match (self) {
        A{ => 1,
        B => 2,
    };
}

fun main() {}
```

Output:
```
zsh: segmentation fault  ./tolk/tolk crash_ast.h_203_17.tolk
```

Backtrace:
```
* thread #1, name = 'tolk', stop reason = signal SIGSEGV: address not mapped to object (fault address: 0x0)
  * frame #0: 0x00005555555a795c tolk`tolk::AssignSymInsideFunctionVisitor::visit(tolk::Vertex<(tolk::ASTNodeKind)37> const*) [inlined] tolk::Vertex<(tolk::ASTNodeKind)1> const* tolk::ASTNodeBase::try_as<(tolk::ASTNodeKind)1>(this=0x0000000000000000) const at ast.h:203:17
    frame #1: 0x00005555555a795c tolk`tolk::AssignSymInsideFunctionVisitor::visit(this=0x00007fffffffab70, v=0x0000555555b058b0) at pipe-resolve-identifiers.cpp:205:54
    frame #2: 0x00005555555a772c tolk`tolk::AssignSymInsideFunctionVisitor::visit(tolk::Vertex<(tolk::ASTNodeKind)36> const*) [inlined] tolk::ASTVisitor::visit_children(this=0x00007fffffffab70, v=0x0000555555b05b30) at ast-visitor.h:65:7
    frame #3: 0x00005555555a76eb tolk`tolk::AssignSymInsideFunctionVisitor::visit(tolk::Vertex<(tolk::ASTNodeKind)36> const*) [inlined] tolk::ASTVisitorFunctionBody::visit(this=0x00007fffffffab70, v=0x0000555555b05b30) at ast-visitor.h:132:67
    frame #4: 0x00005555555a76eb tolk`tolk::AssignSymInsideFunctionVisitor::visit(this=0x00007fffffffab70, v=0x0000555555b05b30) at pipe-resolve-identifiers.cpp:194:13
```

### LINKED ISSUE #1832: Tolk: `pipe-register-symbols`: Crash on invalid struct opcode
MRE:
```
struct(0x2345666666666666666666666666666666666666666666666666666666666666666666666666789) S {}
```

Output:
```
[ 0][t 0][2025-10-11 06:41:45.303444804][refcnt.hpp:288]        Check `ptr && "deferencing null Ref"` failed
zsh: IOT instruction  ./tolk/tolk crashes/crash_check_28_3.tolk
```

Backtrace:
```
    frame #3: 0x00005555557cc180 tolk`td::process_fatal_error(message=<unavailable>) at logging.cpp:287:3
    frame #4: 0x00005555557cdc3d tolk`td::DefaultLog::append(this=<unavailable>, slice=CSlice @ 0x0000562e13e58500, log_level=0) at logging.cpp:265:7
    frame #5: 0x00005555557cb004 tolk`td::Logger::~Logger(this=0x00007fffffffa838) at logging.cpp:130:10
    frame #6: 0x00005555557bd8af tolk`td::detail::process_check_error(message="ptr && \"deferencing null Ref\"", file="<edited>/ton/crypto/common/refcnt.hpp", line=288) at check.cpp:28:3
    frame #7: 0x000055555579e5c8 tolk`td::cmp(td::Ref<td::Cnt<td::BigIntG<257, td::BigIntInfo>>>, long long) [inlined] td::Ref<td::Cnt<td::BigIntG<257, td::BigIntInfo>>>::operator->(this=<unavailable>) const at refcnt.hpp:288:7
    frame #8: 0x000055555579e594 tolk`td::cmp(x=td::RefInt256 @ 0x00007fffffffaa08, y=0) at refint.cpp:257:10
    frame #9: 0x000055555559f308 tolk`tolk::register_struct(tolk::Vertex<(tolk::ASTNodeKind)65> const*, tolk::StructData const*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>>, tolk::GenericsSubstitutions const*) [inlined] bool td::operator<<int>(x=td::RefInt256 @ 0x00007fffffffa980, y=0) at refint.h:89:10
    frame #10: 0x000055555559f2ce tolk`tolk::register_struct(v=0x0000555555b05100, base_struct_ref=0x0000000000000000, override_name=<unavailable>, substitutedTs=0x0000000000000000) at pipe-register-symbols.cpp:150:26
```

### LINKED ISSUE #1833: Tolk: Optimizer: Crash on `throw` argument
MRE:
```
fun t(): never { throw 1222222222222222222222222222222223; }
fun main() { __expect_type(t(), "never"); }
```

Output:
```
"Asm.fif" include
// automatically generated from crash_optimize.cpp_155_20.tolk
PROGRAM{
  0 DECLMETHOD main()
  // 1 functions inlined in-place:
  // - t (1 call)
  // crash_optimize.cpp_155_20.tolk:2
terminate called after throwing an instance of 'std::out_of_range'
  what():  stoul
```

Backtrace:
```
    frame #8: 0x0000555555763d2a tolk`tolk::Optimizer::detect_rewrite_big_THROW() [inlined] unsigned long __gnu_cxx::__stoa<unsigned long, unsigned long, char, int>(__convf=<unavailable>, __name=<unavailable>, __str="1222222222222222222222222222222223", __idx=0x0000000000000000, __base=10) at string_conversions.h:88:2
    frame #9: 0x0000555555763cff tolk`tolk::Optimizer::detect_rewrite_big_THROW() [inlined] std::__cxx11::stoul(__str="1222222222222222222222222222222223", __idx=0x0000000000000000, __base=10) at basic_string.h:4175:12
    frame #10: 0x0000555555763cff tolk`tolk::Optimizer::detect_rewrite_big_THROW(this=0x00007fffffff9b50) at optimize.cpp:155:20
    frame #11: 0x000055555577672d tolk`tolk::Optimizer::find_at_least(this=0x00007fffffff9b50, pb=1) at optimize.cpp:843:10
    frame #12: 0x000055555577976b tolk`tolk::Optimizer::optimize() [inlined] tolk::Optimizer::find(this=0x00007fffffff9b50) at optimize.cpp:883:9
    frame #13: 0x000055555577973f tolk`tolk::Optimizer::optimize(this=0x00007fffffff9b50) at optimize.cpp:892:10
```

### LINKED ISSUE #1834: Tolk: Typechecker: Crash on unexpected argument of `__expect_type`
MRE:
```
fun main() { __expect_type([1, ], "[int]"!); }
```

Output:
```
terminate called after throwing an instance of 'std::bad_alloc'
  what():  std::bad_alloc
zsh: IOT instruction  ./tolk/tolk crashes/crash_pipe-check-inferred-types.cpp_36_10.tolk
```

Backtrace:
```
    frame #9: 0x00005555555e609c tolk`tolk::to_string[abi:cxx11](std::basic_string_view<char, std::char_traits<char>>) [inlined] std::allocator<char>::allocate(this=0x00007fffffffa958, __n=<unavailable>) at allocator.h:196:32
    frame #10: 0x00005555555e609c tolk`tolk::to_string[abi:cxx11](std::basic_string_view<char, std::char_traits<char>>) [inlined] std::allocator_traits<std::allocator<char>>::allocate(__a=0x00007fffffffa958, __n=<unavailable>) at alloc_traits.h:515:20
    frame #11: 0x00005555555e609c tolk`tolk::to_string[abi:cxx11](std::basic_string_view<char, std::char_traits<char>>) [inlined] std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>>::_S_allocate(__a=0x00007fffffffa958, __n=<unavailable>) at basic_string.h:131:16
    frame #12: 0x00005555555e609c tolk`tolk::to_string[abi:cxx11](std::basic_string_view<char, std::char_traits<char>>) [inlined] std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>>::_M_create(this="0\xab\xff\xff\xff\U0000007f"..., __capacity=<unavailable>, __old_capacity=0) at basic_string.tcc:159:14
    frame #13: 0x00005555555e6093 tolk`tolk::to_string[abi:cxx11](std::basic_string_view<char, std::char_traits<char>>) [inlined] void std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>>::_M_construct<char const*>(this="0\xab\xff\xff\xff\U0000007f"..., __beg="", __end="", (null)=<unavailable>) at basic_string.tcc:229:14
    frame #14: 0x00005555555e608d tolk`tolk::to_string[abi:cxx11](std::basic_string_view<char, std::char_traits<char>>) [inlined] std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>>::basic_string(this="0\xab\xff\xff\xff\U0000007f"..., __s="", __n=93824998200144, __a=<unavailable>) at basic_string.h:632:2
    frame #15: 0x00005555555e606f tolk`tolk::to_string[abi:cxx11](std::basic_string_view<char, std::char_traits<char>>) [inlined] std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>>::basic_string(this="0\xab\xff\xff\xff\U0000007f"..., __svw=__sv_wrapper @ 0x00005567e581e690, __a=<unavailable>) at basic_string.h:182:9
    frame #16: 0x00005555555e606f tolk`tolk::to_string[abi:cxx11](std::basic_string_view<char, std::char_traits<char>>) [inlined] std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>>::basic_string<std::basic_string_view<char, std::char_traits<char>>, void>(this="0\xab\xff\xff\xff\U0000007f"..., __t=<unavailable>, __a=<unavailable>) at basic_string.h:801:4
    frame #17: 0x00005555555e606f tolk`tolk::to_string[abi:cxx11](string_view=(_M_len = 93824998200144, _M_str = "")) at pipe-check-inferred-types.cpp:36:10
    frame #18: 0x0000555555571257 tolk`tolk::handle_possible_compiler_internal_call(cur_f=0x0000555555b15230, v=<unavailable>) at pipe-check-inferred-types.cpp:158:64
```

### LINKED ISSUE #1835: Tolk: Typechecker: Crash on unexpected `.storeSlice` argument
MRE:
```tolk
fun main() {
  var test = builder.storeSlice(1, 5);
}
```

Output:
```
zsh: segmentation fault  ./tolk/tolk crash_pipe-infer-types-and-calls.cpp_1248_29.tolk
```

Backtrace:
```
* thread #1, name = 'tolk', stop reason = signal SIGSEGV: address not mapped to object (fault address: 0x10)
  * frame #0: 0x00005555555cb1b3 tolk`tolk::InferTypesAndCallsAndFieldsVisitor::infer_function_call(this=0x00007fffffffaaf0, v=0x0000555555b05260, flow=0x00007fffffffa968, used_as_condition=false, hint=0x0000555555ab9760) at pipe-infer-types-and-calls.cpp:1248:29
    frame #1: 0x00005555555bde20 tolk`tolk::InferTypesAndCallsAndFieldsVisitor::infer_any_expr(this=<unavailable>, v=<unavailable>, flow=<unavailable>, used_as_condition=<unavailable>, hint=<unavailable>) at pipe-infer-types-and-calls.cpp:390:16
    frame #2: 0x00005555555beaea tolk`tolk::InferTypesAndCallsAndFieldsVisitor::infer_assignment(this=0x00007fffffffaaf0, v=0x0000555555b054b0, flow=0x00007fffffffa968, used_as_condition=false) at pipe-infer-types-and-calls.cpp:471:12
    frame #3: 0x00005555555bdfbe tolk`tolk::InferTypesAndCallsAndFieldsVisitor::infer_any_expr(this=<unavailable>, v=<unavailable>, flow=<unavailable>, used_as_condition=<unavailable>, hint=0x0000000000000000) at pipe-infer-types-and-calls.cpp:364:16
```

### LINKED ISSUE #1837: Tolk: Logger crash on invalid asm constant
MRE:
```tolk
fun t2(): int asm "90()437508 PUSHINT"""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""" }
fun t2() {
    var stateInitCe""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""!""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""try"""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""createMessage""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""	"""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""ÿ"""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""?""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""ÿ""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""D""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""";
```

Output:
```
[ 0][t 0][2025-10-11 06:05:08.209326864][refcnt.hpp:288]        Check `ptr && "deferencing null Ref"` failed
zsh: IOT instruction  ./tolk/tolk crashes/crash_logging.cpp_287_3.tolk
```

Backtrace:
```
    frame #3: 0x00005555557cc180 tolk`td::process_fatal_error(message=<unavailable>) at logging.cpp:287:3
    frame #4: 0x00005555557cdc3d tolk`td::DefaultLog::append(this=<unavailable>, slice=CSlice @ 0x00005653e129a800, log_level=0) at logging.cpp:265:7
    frame #5: 0x00005555557cb004 tolk`td::Logger::~Logger(this=0x00007fffffffa778) at logging.cpp:130:10
    frame #6: 0x00005555557bd8af tolk`td::detail::process_check_error(message="ptr && \"deferencing null Ref\"", file="<edited>/ton/crypto/common/refcnt.hpp", line=288) at check.cpp:28:3
    frame #7: 0x00005555556a738d tolk`tolk::AsmOp::IntConst(tolk::SrcLocation, td::Ref<td::Cnt<td::BigIntG<257, td::BigIntInfo>>> const&) [inlined] td::Ref<td::Cnt<td::BigIntG<257, td::BigIntInfo>>>::operator->(this=<unavailable>) const at refcnt.hpp:288:7
    frame #8: 0x00005555556a7375 tolk`tolk::AsmOp::IntConst(loc=(file_id = -1, char_offset = -1), x=0x00007fffffffa980) at asmops.cpp:0
    frame #9: 0x00005555556a8138 tolk`tolk::AsmOp::Parse(loc=(file_id = -1, char_offset = -1), custom_op="90()437508 PUSHINT") at asmops.cpp:227:12
    frame #10: 0x00005555556a8a7e tolk`tolk::AsmOp::Parse(loc=<unavailable>, custom_op=<unavailable>, args=0, retv=<unavailable>) at asmops.cpp:234:14
    frame #11: 0x00005555556451fd tolk`tolk::ConvertASTToLegacyOpVisitor::start_visiting_function(tolk::FunctionData const*, tolk::Vertex<(tolk::ASTNodeKind)59> const*) [inlined] tolk::convert_asm_body_to_AsmOp(fun_ref=<unavailable>, asm_body=0x0000555555b1aee0) at pipe-ast-to-legacy.cpp:2380:25
    frame #12: 0x0000555555644657 tolk`tolk::ConvertASTToLegacyOpVisitor::start_visiting_function(fun_ref=<unavailable>, (null)=<unavailable>) at pipe-ast-to-legacy.cpp:2450:7
    frame #13: 0x000055555562c136 tolk`tolk::pipeline_convert_ast_to_legacy_Expr_Op() [inlined] void tolk::visit_ast_of_all_functions<tolk::ConvertASTToLegacyOpVisitor>() at ast-visitor.h:225:7
```

### LINKED ISSUE #1838: Tolk: Analyzer: Crash on large bigint
MRE:
```
fun main() { 99999999999999999999999899199999~/340282366920838463463374607431769011456; }
```

Output:
```
"Asm.fif" include
// automatically generated from crash_analyzer.cpp_842_18.tolk
PROGRAM{
  0 DECLMETHOD main()
  DECLPROC test7()
zsh: segmentation fault  ./tolk/tolk crashes/crash_analyzer.cpp_842_18.tolk
```

### LINKED ISSUE #1839: Match statement falls through on invalid enum values in Tolk 1.1
# Background

While writing tests for the TEP-524 vault standard, I discovered that Tolk's `enum` with `match` statement exhibits abnormal behavior: when an enum value that doesn't match any branch is passed, instead of throwing an error, the program incorrectly enters the last branch.

Full code reference: https://github.com/ipromise2324/tolk-enum-match-bug

# Description

## Test Code

First, I defined a `RoundingType` enum and a function using `match`:

Image

### Trigger Method

An **invalid** enum value is passed through an internal message:

Image

## Expected Behavior

The value of `wrongRoundingType` is `5`, which is outside the valid range of `RoundingType` (0, 1, 2). Therefore, the `match` statement should not enter any branch and should proceed to execute `throw 1000`.

## Actual Behavior

Execution output shows:

Image

Observations:

- ✅ The value of `rounding` is indeed `5`
- ✅ The value of `RoundingType.HalfUp` is indeed `2`
- ✅ The result of `rounding == RoundingType.HalfUp` is `0` (false)
- ❌ **However, the program entered the `HalfUp` branch** instead of throwing an exception

This indicates that the `match` statement, when handling out-of-range enum values, incorrectly enters the last branch instead of continuing to execute subsequent code.

# Environment

## Dependencies

```json
{
    "@ton/blueprint": ">=0.40.0",
    "@ton/sandbox": ">=0.37.0",
    "@ton/test-utils": ">=0.11.0",
    "@types/jest": "^30.0.0",
    "@types/node": "^22.17.2",
    "jest": "^30.0.5",
    "prettier": "^3.6.2",
    "@ton/ton": ">=15.3.1 <16.0.0",
    "@ton/crypto": "^3.3.0",
    "ts-jest": "^29.4.1",
    "ts-node": "^10.9.2",
    "typescript": "^5.9.2",
    "@ton/tolk-js": ">=1.0.1",
    "@tact-lang/compiler": ">=1.6.13 <2.0.0",
    "@ton-community/func-js": ">=0.10.0"
}
```

## Hardware

- **OS**: macOS
- **Device**: MacBook Pro M4 (2024)

### LINKED ISSUE #1864: Tolk: never value can be obtained by parsing a slice
The following code:

```tolk
tolk 1.1

struct(0b0) DirectContent<T> {
    value: T
}
struct(0b1) RefContent<T> {
    value: Cell<T>
}
type EitherContent<T> = DirectContent<T> | RefContent<T>;

fun EitherContent<T>.load(self): T {
    match (self) {
        DirectContent => { return self.value; }
        RefContent    => { return self.value.load(); }
    }
}

struct(0x73629999) InCall<T> {
    forward: EitherContent<T>
}



fun onInternalMessage(in: InMessage) {
    val msg = InCall<never>.fromSlice(in.body);
    val nev: never = msg.forward.load();
    throw 24;
}
```
has exit code 0 on message `pytoniq.begin_cell().store_uint(0x73629999,32).store_uint(0,1).end_cell().to_boc().hex()` (`b5ee9c720101010100070000097362999940`).

It should not do that, by either throwing that invalid value is deserialized or throwing 24. Instead, the slice is checked for being empty and parsing succeeds, and back to `onInternalMessage` the variable `nev` is treated as uninhabited dropping all subsequent code.

This is different from #1839 because this does not use any unsafe primitives (asm, `as` casts).

tolk-vm:
This release mainly focuses on TVM 12 upgrade, which introduces new assembler instructions to make smart contract development easier and cheaper. In addition, the compiler has been enhanced with several low-level capabilities, undefined-behavior detection, and various minor bug fixes.

1. Breaking change: `address` is now "internal only"
2. Rich bounces: not 256 bits, but the full body on bounce
3. Cheap builder-to-slice, StateInit, and address composition
4. Compilation errors significantly improved
5. Anonymous functions supported
6. Borrow checker and other stuff related to undefined behavior
7. Several low-level capabilities and minor fixes/improvements

A lot of text below. Ready to dig into?

# Breaking change: `address` is now "internal only"

Before:
* `address` meant internal/external/none

Now:
* `address` — internal only
* `address?` (nullable) — internal/none, exactly like "maybe address" in `@ton/core`
* `any_address` — internal/external/none

### Motivation to change `address` behavior

While developing smart contracts, **in 90% cases, you need only internal addresses**. In 9% cases, you need "none" — to describe "the absence of an address". In 1% cases, you need external addresses — to emit logs or write a chain proxy.

Having a struct
```
struct IncomingMsg {
    owner: address
}
```
the developer must remember: a **none/external** address may occur here, and it must be validated via `msg.owner.isInternal()` before handling it.

However, validation is not always mandatory:
- for untrusted input — recommended
- for trusted input (e.g. from a storage that you've saved by yourself) — no need, to avoid wasting gas
- sometimes "none" address is okay (e.g., "address is not yet filled" in a storage), then you use `isInternal()/isNone()` to distinguish them

As a result, **contract's code is cluttered with isInternal()**, which can be easily forgotten. 

In TL/B, there is `MsgAddress` and `MsgAddressInt`. In Tolk — only `address`. It's **the only place where Tolk's type system is less expressive** compared to TL/B.

Moreover: in a js library `@ton/core` there are distinct TypeScript classes `Address` and `ExternalAddress`. The last is almost unused, but the first exists literally everywhere. But **while auto-generating TypeScript wrappers for contracts** <u>it's incorrect</u> to map `address` to js `Address` — because of none/external. Addresses are the most painful part for ABI and wrapper generation I'm currently working on. Absolutely as expected.

### Possible solution for `address`

1) Do nothing — leave `address` as internal/external/none. In TypeScript, convert `address` **into a union** `Address | ExternalAddress | NoneAddress`. In 90% cases the developer should manually do `instanceof Address` or unsafe `as Address` to pass it into client code. It will be painful.

2) Modify `@ton/core`, so that `Address` could also be internal/external/none. Extend it by methods similar to stdlib: `isInternal()`, `getWorkchain()`, etc. In brief, **make an address on a client equal to an address in a language**. I don't consider this scenario bad — it's also available. Depends on whether we want the user to always check `isInternal()` on a client or not.

3) Modify Tolk, to separate "internal address" from other varieties. In other words — **make an address in a language equal to an address on a client**, having greatly simplified smart contracts development and match TL/B: you exactly see where an address is internal, where not, you don't need `isInternal()` checks, it becomes compatible with `@ton/core` and expressive in the future ABI, etc.

### Why "case 3" was not implemented before

A reasonable question: if "just address" has such drawbacks, why in Tolk v1.0 it meant "anything", why addresses were not distinct initially?

The answer: **because TVM knows nothing about addresses**. We need auto-serialization, but TVM has only one instruction `LDMSGADDR` — parse any valid TL/B address. No way to load and validate "internal only" or a combination.

Hence, if we wanted `address` to be only internal — the compiler should have inserted extra validation which would **significantly increase gas cost**. Absolutely everywhere — for example, addresses in a storage would also require additional gas. As a result, automatic serialization would be much less effective than manual when you just don't check anything.

If we want to make addresses distinct, it should be done at the TVM level. 

**TVM 12 has new instructions for working with internal/none addresses**. They are made specially to support `address` auto-validated on (de)serialization, and `address?` (null/slice on a stack) serialized as internal/none and also auto-validated.

### `address`, `address?`, and `any_address`

Since Tolk v1.2, addresses will work the following way:
* `address` — internal only
* `address?` (nullable) — internal/none
* `any_address` — internal/external/none

**`address` is automatically checked on (de)serialization** by a new TVM 12 instruction (works like LDMSGADDR + throw 9 if not internal). So, you have a full guarantee that it's definitely an internal (standard) address with workchain and hash. Moreover, `isInternal()` and other methods don't exist for `address`, they make no sense.

**`address?` (nullable) is null/slice on a stack, but serialized as none/internal (2/267 bits)**. In TON, there is a standard from its very beginning: "absence of address" is encoded as "addr_none" ('00' — two zero bits). Such encoding exists in TEPs, it's hardcoded in clients (JS and others). That's why I made a decision that `null` will be serialized as '00' here, equal to "maybe address" in TypeScript. TVM 12 has dedicated instructions to check and (de)serialize "internal or none". _If somebody wants to make a "nullable serialization" (not '00'/address (2/267 bits) but '0'/'1'+address (1/268 bits), he can express it with a custom `Maybe<T>` instead of `address?`)_

**`any_address`** is a new type that behaves like the current `address`. It has methods `isInternal()` and others, as well as `getWorkchain()` applicable if "any" holds internal. 

**If I was smarter, in Tolk v1.0 I'd have made `any_address` instead of `address`**, and right now a new type `address` would have introduced. But I was not that smart, so now we have a breaking change. A breaking change that you'll love.

I'll also highlight that **`address?` is NOT TL/B `(Maybe MsgAddressInt)`**, it's `MsgAddress` with two valid values backed by one stack slot. It cannot be expressed with TL/B. It's one more reason that TL/B and type system are different, and it's incorrect to describe ABI with TL/B.

### How to transition existing contracts to modern addresses?

In other words, what is expected to break when switching to a new compiler?

* if a contract does not use "none" addresses — nothing, since everywhere is internal already
* if a contract uses "none" addresses, when send from a client and tried to parse as `address`, parsing will throw 9, the hardest part is finding where exactly, by exploring TVM logs for example
    * replace `address` with `address?` to accept both internal/none
    * check not `isInternal()/isNone()` but for `null`
* if a contract already uses `address?` (very unlikely), it will be parsed not 1/268 bits but 2/267 bits (none/internal)
* methods `isInternal()` etc. do not exist on type `address`, they exist on type `any_address`

All in all, these changes are logical and expected. Where "none" is awaited — replace `address` with `address?`. Where only internal — most cases — no changes. Having your contract covered with tests, everything is fixed quite easily.

**The best way to migrate an existing contract:**

1. Make it compile. Since `isInternal()` no longer exists for `address`, those calls will fail with a compilation error. Probably, you need `address?` there, and check for `null`. Sometimes, such calls may be dropped if failing at deserialization is suitable.
2. When it compiles but crashes on running a test, it probably crashes with excno 9 while deserializing `address`. Find the location and replace with `address?`

I intentionally don't mention `any_address`, because it's very-very unlikely you really need it.

### What about gas prices of auto-validation internal/none?

TVM 12 has a special "load-and-validate internal address" instruction — `LDSTDADDR`. Its cost is exactly the same as `LDMSGADDR` that validates nothing. Same for loading `address?` supporting "none" and putting `null` on a stack.

So, **during load (deserialization), `address` and `any_address` are absolutely the same** (26 gas).

Moreover, `address` is checked on saving (on serializing) also. In case you've occasionally converted a builder/slice to invalid `address`, it would fire a serialization error (similar to "8 STU" when an integer doesn't fit 8 bits). The "validate-and-store internal address" instruction — `STSTDADDR` — also costs 26 gas, but raw `STSLICE` has a short opcode and costs 18 gas.

So, **during store (serialization), `address` costs a bit more** (26 gas vs 18). It's negligible in practice.

Since addresses are validated by TVM, you have a full guarantee from the type system, so manual checks are no longer required (or forgotten).

# TVM 12: rich bounced messages

Historically, working with bounced messages in TON was cumbersome. The most painful aspect was that when a message is bounced, only the first 256 bits of original message body are returned. It means that you should tune layout of an outgoing message, so that if it's bounced, retrieve necessary info from its first bits. That was not always possible: for instance, an address is 267 bits, so you cannot even decode an address back.

TVM 12 introduces "rich bounced messages": **the entire body can be returned, not first 256 bits**. This mode is more expensive in case of bounce, but greatly simplifies contract logic and architecture. 

More precisely, two new modes were introduced:
- return full body (a tree of cells)
- return the root cell without references (cheaper and often sufficient)

Besides the larger body, you can get exit code, gas used, etc. of a message processed.

**Changes in TVM 12 are reflected in `createMessage` and `onBouncedMessage`**.

### How to send a message with "rich bounce" mode

Before this update, `bounce` parameter was just true or false:
```
// Tolk v1.1 and below
createMessage({
    bounce: true,
    ...
})
```

Now, `bounce` is an enum:
```
createMessage({
    bounce: BounceMode.RichBounce,
    ...
})
```

`BounceMode` enumerates all possible behaviors:
```
enum BounceMode {
    NoBounce
    Only256BitsOfBody       // old way (still exists, cheapest)
    RichBounce              // parse it with `RichBounceBody.fromSlice(in.bouncedBody)`
    RichBounceOnlyRootCell  // same, but `originalBody` will contain only a root cell
}
```

So, instead of true/false, use an enum. It supports both old cheap and new rich bounces.

**true/false in `bounce` is still supported for compatibility**. `false` is `NoBounce`, `true` is `Only256BitsOfBody`. It's expressed with a union:
```
@deprecated
type OldBounceMode = bool

struct CreateMessageOptions<TBody> {
    bounce: BounceMode | OldBounceMode
    ...
}
```

### How to handle a rich bounced message

In `onBouncedMessage`, use `in.bouncedBody`
- sent with `Only256BitsOfBody` — it's "0xFFFFFFFF" and first 256 bits of outgoing body
- sent with `RichBounce` — use `lazy RichBounceBody.fromSlice(in.bouncedBody)`
- sent with `RichBounceOnlyRootCell` — same, but `originalBody` will contain only a root cell

**It's preferred that in `createMessage` you use either old or new everywhere**. If you mix them (some messages are sent in an old way, some in a new way), it becomes not straightforward to handle. Still possible, but you manually need to check incoming prefixes.

**If you use old bounces**, nothing changes for you:
```
fun onBouncedMessage(in: InMessageBounced) {
    // in.bouncedBody is populated depending on BounceMode
    // for instance, "Only256BitsOfBody" means 
    // "bounced prefix 0xFFFFFFFF" and 256 bits of data
    in.bouncedBody.skipBouncedPrefix();

    // read opcode, or use lazy match, remember the 256-bit limit
    val originalOpcode = in.bouncedBody.loadUint(32);
    ...
}
```

**If you use rich bounces**, here is the way:
```
fun onBouncedMessage(in: InMessageBounced) {
    // suppose you ALWAYS use BounceMode.RichBounce in createMessage
    // so you parse ALL bounces as follows
    val rich = lazy RichBounceBody.fromSlice(in.bouncedBody);

    // handle rich.originalBody, probably with lazy match
    // use rich.xxx to get exitCode, gasUsed, and so on
}
```

Besides `originalBody`, you can access some properties of an outgoing message, exit code, etc.
```
struct (0xfffffffe) RichBounceBody {
    // a cell with original body, but for `RichBounceOnlyRootCell` without refs (cheaper)
    originalBody: cell
    // some fields of an original (outgoing) message
    originalInfo: Cell<RichBounceOriginalMsgInfo>
    // see docs
    bouncedByPhase: uint8
    // exception code from `throw` or TVM internals
    exitCode: int32
    // filled if compute phase executed (gasUsed and vmSteps)
    computePhase: RichBounceComputePhaseInfo?
}
```

### Conclusion about bounced messages

- instead of true/false, use `BounceMode.xxx`
- cheap bounces (with 256 bits of body) still exist, and rich bounces (with full body and extra info) now available
- use either always cheap or always rich in a single contract — and handle `in.bouncedBody` correctly in `onBouncedMessage`

# TVM 12: cheap builder-to-slice and address composition

In TON, all data is stored in cells. Cells "opened for reading" are called slices. Cells "at the moment of writing" are called builders. Having a builder, you can only write, not read. Having a slice, you can only read, not write. There is no dedicated entity "address": addresses are just slices, just binary data.

If you have written something to a builder, the only way to get binary data was "builder-to-cell" and then "cell-to-slice". But it's very expensive (500 gas!), because cell creation is expensive. **In TVM 12, there is a cheap instruction `BTOS` — "builder-to-slice"** without intermediate cell creation.

**All in all, it results in**
- `b.endCell().beginParse()` is now very cheap: the compiler automatically replaces it to `b.toSlice()`
- easily convert `builder` to `address`: the same `BTOS`; it means, that hacks with "return a builder with a valid address" can now be removed
- hash of `StateInit` when deploying new contracts can be calculated cheaper

### Changes in stdlib

- new method: `builder.toSlice` (asm `BTOS`)
- new method: `builder.hash` (asm `HASHBU`)
- `address.buildSameAddressInAnotherShard` deprecated, use `address.calculateSameAddressInAnotherShard`
- `AutoDeployAddress.buildAddress` deprecated, use `AutoDeployAddress.calculateAddress`
- `AutoDeployAddress.addressMatches` deprecated, use `AutoDeployAddress.calculateAddress` and operator `==`

### Compiler optimizations

Even "old-fashioned" code now works more efficiently:
```
// automatically replaced by `x{01} PUSHSLICE`
beginCell().storeInt(1, 8).endCell().beginParse()

// automatically replaced by `BTOS`
b.endCell().beginParse()

// automatically replaced by `<b b> PUSHREF`
beginCell().endCell()
```

More precisely, the following peephole optimizations introduced:
- ENDC + CTOS => BTOS
- ENDC + HASHCU => HASHBU
- NEWC + x{...} STSLICECONST + BTOS => x{...} PUSHSLICE
- NEWC + ENDC + CTOS => createEmptySlice
- NEWC + ENDC => createEmptyCell

Besides them, calculating addresses for `createMessage()` when using `stateInit` and `toShard` were optimized with `BTOS`. **So, simply switching to TVM 12 and Tolk v1.2 will reduce gas usage in jettons and other contracts.**

# Compilation errors improved

Prior to this release, when a compilation error occurred, the compiler showed description and "some location".

Now, instead of showing just "some location", errors are underlined and formatted nicely:

```
dev.tolk:11:15: error: can not assign `Point` to variable of type `int`

    // in function `main`
  11 |     counter = point; 
     |               ^^^^^
```

```
dev.tolk:7:12: warning: variable `counter` of type `int` can never be `null`, this condition is always false

    // in function `main`
   7 |     while (counter == null) {
     |            ^^^^^^^^^^^^^^^
```

```
dev.tolk:6:35: error: method `loadUint` not found for type `builder`
           (but it exists for type `slice`)

    // in function `main`
   6 |     beginCell().storeUint(123, 8).loadUint();
     |                                   ^^^^^^^^
```

```
dev.tolk:6:12: error: can not convert type `cell` to return type `int`

    // in function `main`
   6 |     return beginCell()
     |            ^^^^^^^^^^
 ...   ...
  10 |         .endCell();
     |         ^^^^^^^^^^
```

Internally, every AST vertex now has a range (from/to offsets). Those ranges are precicely calculated while building AST from tokens, accurately dealing with comments, newlines, etc. Every assembler instruction in the Fift output now has an "origin" AST vertex where it comes from. Therefore, line comments in Fift output also became more precise.

In the future, AST ranges will be part of source mapping (enabling debugging, stacktrace mapping at runtime, and many awaited features).

# Anonymous functions supported

Now you can use lambdas — function expressions, without capturing outer variables. They can be passed as callbacks, assigned to variables, etc.

```
fun callAndAssertTrue(callback: () -> bool) {
    val result = callback();
    assert (result == true) throw 123;
}

callAndAssertTrue(fun() {   // lambda
    ...
    return ...;
});
```

Unlike regular functions, where you must specify all parameters types, lambda parameters types may be omitted if they can be inferred:
```
fun Result.pushMath(mutate self, loadFn: (int, slice) -> int) {
    val ans = mathFn(32, "...");
    self.output.push(ans);
}

r.pushMath(fun(bits, s) {    // bits is `int`, s is `slice`
    return s.loadUint(bits)
});

// but it's an error: param's type cannot be inferred here:
val doubleFn = fun(param) { return param * 2 };
// correct is:
val doubleFn = fun(param: int) { return param * 2 };
```

From the type system point of view, a function (and a lambda) has a special type `(...ArgsT) -> ReturnT`.

```
// a regular function
fun abs(a: int): int { ... }

val cb1 = abs;                           // (int) -> int
val cb2 = fun(a: coins): MyData { ... }; // (coins) -> MyData 
val cb3 = fun(a: slice) {};              // (slice) -> void
```

As first-class functions, lambdas can even be returned:
```
fun createFinalizer() {
    return fun(b: builder) {
        b.storeUint(0xFFFFFFFF, 32);
        return b.toSlice();
    }
}

val f = createFinalizer();    // (builder) -> slice
f(beginCell());               // slice with 32 bits
```

While lambdas are not common in smart contracts, they become useful in general purpose tools. They can easily be combined with generics of any level, nested into each other, and so on.

Note that lambdas are not closures: capturing outer variables not supported.
```
fun outer(x: int) {
    return fun(y: int) {
        return x + y;    // error: undefined symbol `x`
    }
}
```

Capturing variables is nearly impossible to implement on a stack machine, just like inheritance (conceptually equivalent).

# UB (undefined behavior) prevention in some cases

Undefined behavior is a thing that "always exists" to some extent in low-level languages. Sutiation when a program behaves not as you expect — but it's not a bug, because multiple behaviors are correct simultaneously.

Let's provide a simple example:
```
var x = 100;
x += (x = 0);
// 0 or 100?
```
Some of you will say "x will be 0"! Some will respond "x will be 100"! And no one is wrong, as well is no one is correct. It's _undefined_, both results are valid. Producing either 0 or 100 — the compiler is correct in both scenarios.

From one point of view, the right part (x=0) should be evaluated first, then "(x already 0) += 0" would be 0. From the other, "x += yyy" equal to "x = x + yyy", so it becomes "x = 100 + (x=0)", resulting in 100.

We can even extend the example:
```
fun int.inc(mutate self, dx: int) { self += dx }
fun int.del(mutate self) { self = 0; return 0 }

var x = 100;
x.del(); x.inc(x);   // definitely 0
x.inc(x.del());      // 0 or 100?
```

Again, neither 0 nor 100 is correct, and neither is wrong. This snippet is absolutely equal to the first one, and if the answers seem different to you, congratulations: you've just faced *the undefined*.

Since Tolk v1.2, the compiler gives a compilation error in such cases.

### Mini borrow checker for UB prevention

When concurrent mutations of the same memory piece occur, the compiler gives an error:
```
struct Items { x: map<...>, y: map<...> }
// and methods addX/addY/removeX/removeY

items.addY(items.removeX(1), true);
```

now result in
```
dev.tolk:27:16: error: can not borrow `items` for mutation once again, it is already being mutated by `Items.addY`
           hint: split a complex expression into several simple ones

    // in function `test_dict`
  27 |     items.addY(items.removeX(1), true);
     |                ^^^^^
```

Other prohibited examples:
* x += (x = 0)
* x.add(x.del())
* f(mutate x, mutate x)
* d1.assign(someCall(mutate d1))

The compiler analyzes not only variables as a whole, but particular indices of an object/tensor. Independent mutations are okay, but when fields overlap, it's an error:
```
struct Point { x: int, y: int }
fun Point.annulate(mutate self) { ... }

// error, can not borrow `p`, because `p.x` already borrowed
p.x += p.annulate().y

// error, can not borrow `d.nested.field`, because `d` already borrowed
d.mutatingMethod(d.nested.field = 10)
```

Even though writing such a code is impractical, if you occasionally do this, you'd waste a lot of time understanding what's going wrong. Now such code is prohibited, which makes development much safer.

### Preventing runtime UB with enums on intentional stack corruption

There is a recently reported bug (https://github.com/ton-blockchain/ton/issues/1839) that when a stack is corrupted and contains an apriori invalid value, the compiler enters a wrong `match` branch. While this might be shocking at the first glance, actually it's just undefined behavior. An irregular stack value could not occur in normal execution (not via deserialization, not another way). It could occur only if a developer intentionally corrupts the stack via unsafe casts or asm injections.

Unlike compile-time UB that can theoretically be prevented, runtime UB is an orthogonal thing. 

In this exact case, I decided to change the behavior. Every `match` for an enum now contains a hidden "else { throw 5 }" branch, so that even if a stack is corrupted, execution will not enter the last branch. Also, I've reconsidered how exhaustive matching affect control flow.

It does not mean that this solution is better than the previous, no. For instance, it leads to bytecode growth and extra runtime checks. The essence of "unexpected behavior" is that no best solution exists, by definition. None is preferable, and every is correct. But probably, some are not so "unexpected" than the others, so thanks to pointing this out.

Fixes #1839

And keep in mind, that you can always corrupt the stack via asm functions, anyway. For example, union types and pattern matching work via type-id on a stack. If you pollute it intentionally, the execution will evolve unpredictably, this ambiguity is expected. So, this change focuses on enums only.

### Combinations of assembler NOT and xxxNOT instructions

The compiler optimizes common constructions like
```
assert (!isEmpty) throw 123;
```

In assembler, when `isEmpty` variable is at the top of a stack, it looks like
```
NOT
123 THROWIFNOT
```

Obviously, it can be optimized to `123 THROWIF`, but... Generally, it cannot :)

Why? Because `NOT` is a bitwise operation, but `THROWIFNOT` is logical. For booleans (-1 and 0) they are equivalent, but for non-booleans, they are different. For example, having `2` on a stack, `NOT` = ~2 = -3, so `NOT + THROWIFNOT` does not throw, whereas `THROWIF` does.

It's very unlikely you'd use bitwise negation in combination with IF(NOT) or THROWIF(NOT), but anyway, for correctness, such kind of replacements should be done semantically, not as peephole. 

This version of compiler deals with bitwise/logical inversions in combination with logical operators. For instance, the snippet above will be optimized into `123 THROWIF` when `isEmpty` is `bool`, but will not if it's `int`.

Also, the following peephole optimizations are added:
* 0 EQINT + N THROWIF -> N THROWIFNOT and vice versa
* 0 NEQINT + N THROWIF -> N THROWIF
* N EQINT + NOT -> N NEQINT 
* and other mathematical operations + NOT, like `!(a >= 4)` -> `a < 4`

As a result, conditions of `do while` are nicely negated to TVM's UNTIL.

# Low-level capabilities/fixes/improvements

Introduced compile-time constant expressions which are valid for constants / fields / enum initializers; instead of deferring evaluation to the IR, they are evaluated directly on the AST at compile-time, 
* will be compatible with future ABI (it will contain "constant expressions" for field defaults)
* `enum` member initializers can include any complex expressions, including `ton` or `stringCrc32` (earlier they were restricted to simple numbers)

Serialization type `coins` supported for enums.

Custom `packToBuilder` and `unpackFromSlice` may exist independently (earlier they should have been defined together).

`match` statements now support addresses.

Several built-ins added to stdlib. For instance, **`T.toTuple()`** — convert any object to a `tuple` and back.
```
/// Packs any object from a stack to a tuple. 
/// An object occupies N slots on a stack — the tuple will be of size N.
fun T.toTuple(self): tuple

/// Unpacks a tuple back to an object on a stack.
fun T.fromTuple(packedObject: tuple): T

/// Get human-readable type name of any T for logging or low-level purposes.
/// - `int.typeName()` = `42.typeNameOfObject()` = "int"
fun T.typeName(): slice
fun T.typeNameOfObject(self): slice
```

A new file `@stdlib/exotic-cells` with functions of parsing/creating exotic cells and particularly library references.

Fixed `map.mustGet()` with error codes > 2048.

Fixed several minor bugs found by fuzzing. None of them can appear in real contracts, but yes, the compiler could clash on malformed syntax or incorrect Fift code in asm strings trying to parse it. Closes #1830. Closes #1831. Closes #1832. Closes #1833. Closes #1834. Closes #1835. Closes #1837. Closes #1838.

Also, `void` is now used instead of `never` to indicate fields in structs that make no sense and can be missed out. Both these types are empty and mean "absence" or "non-existence", but `never` conflicts with control flow edges. Fixes #1864 — an interesting corner case.

Added LLDB pretty printers for all widely used C++ types. Now debugging compiler's sources becomes much easier.

Screenshot 2025-11-05 at 18 49 10

# Related pull requests

* https://github.com/ton-blockchain/tolk-js/pull/15
* https://github.com/ton-blockchain/tolk-bench/pull/4
* https://github.com/ton-community/ton-docs/pull/1785
* https://github.com/ton-org/docs/pull/1222
* JetBrains IDE: https://github.com/ton-blockchain/intellij-ton/releases/tag/v3.1.0
* VS Code and LSP: https://github.com/ton-blockchain/ton-language-server/releases/tag/v0.5.1

### DIFF
--- a/crypto/smartcont/tolk-stdlib/common.tolk
+++ b/crypto/smartcont/tolk-stdlib/common.tolk
@@ -1,7 +1,7 @@
 // Standard library for Tolk (LGPL licence).
 // It contains common functions that are available out of the box, the user doesn't have to import anything.
 // More specific functions are required to be imported explicitly, like "@stdlib/gas-payments".
-tolk 1.1
+tolk 1.2
 
 /**
     Built-in types.
--- a/crypto/smartcont/tolk-stdlib/exotic-cells.tolk
+++ b/crypto/smartcont/tolk-stdlib/exotic-cells.tolk
@@ -1,0 +1,55 @@
+// A part of standard library for Tolk
+tolk 1.2
+
+enum ExoticCellType: int8 {
+    Ordinary = -1
+    PrunedBranch = 1
+    LibraryReference = 2
+    MerkleProof = 3
+    MerkleUpdate = 4
+}
+
+/// Transforms an ordinary or exotic cell into a slice, as if it were an ordinary cell. 
+/// A flag is returned indicating whether c is exotic. 
+/// If so, ExoticCellType can later be deserialized from the first eight bits of s.
+/// ```
+/// var (s, isExotic) = c.beginParseSpecial();
+/// if (isExotic) {
+///     // use unsafe cast because the first 8 bits hold a correct value, no need for validation
+///     val cellType = s.loadInt(8) as ExoticCellType;
+/// }
+/// ```
+@pure
+fun cell.beginParseSpecial(self): (slice, bool)
+    asm "XCTOS"
+
+/// Creates an exotic or ordinary cell from a builder.
+@pure
+fun builder.endCellSpecial(self, isExotic: bool = true): cell
+    asm "ENDXC"
+
+/// Returns i-th hash of the cell.
+@pure
+fun cell.hashOfLevel(self, level: int): uint256
+    asm "CHASHIX"
+
+/// Returns i-th depth of the cell.
+@pure
+fun cell.depthOfLevel(self, level: int): int
+    asm "CDEPTHIX"
+
+/// Loads an exotic cell and returns an ordinary cell.
+/// If the cell is already ordinary, does nothing. Throws an exception if cannot be loaded.
+@pure
+fun cell.loadSpecial(self): cell
+    asm "XLOAD"
+
+/// Transforms an ordinary or exotic cell into a library reference
+/// (into an exotic cell with contents "8 bits cell type" + "256 bits lib hash").
+@pure
+fun cell.toLibraryReference(self): cell {
+    return beginCell()
+        .storeInt(ExoticCellType.LibraryReference as int, 8)
+        .storeUint(self.hash(), 256)
+        .endCellSpecial()
+}
--- a/crypto/smartcont/tolk-stdlib/gas-payments.tolk
+++ b/crypto/smartcont/tolk-stdlib/gas-payments.tolk
@@ -1,5 +1,5 @@
 // A part of standard library for Tolk
-tolk 1.1
+tolk 1.2
 
 /**
   Gas and payment related primitives.
--- a/crypto/smartcont/tolk-stdlib/lisp-lists.tolk
+++ b/crypto/smartcont/tolk-stdlib/lisp-lists.tolk
@@ -1,5 +1,5 @@
 // A part of standard library for Tolk
-tolk 1.1
+tolk 1.2
 
 /**
     Lisp-style lists are nested 2-elements tuples: `[1, [2, [3, null]]]` represents list `[1, 2, 3]`.
--- a/crypto/smartcont/tolk-stdlib/tvm-dicts.tolk
+++ b/crypto/smartcont/tolk-stdlib/tvm-dicts.tolk
@@ -1,5 +1,5 @@
 // A part of standard library for Tolk
-tolk 1.1
+tolk 1.2
 
 /**
     Low-level API working with TVM dictionaries.
--- a/crypto/smartcont/tolk-stdlib/tvm-lowlevel.tolk
+++ b/crypto/smartcont/tolk-stdlib/tvm-lowlevel.tolk
@@ -1,5 +1,5 @@
 // A part of standard library for Tolk
-tolk 1.1
+tolk 1.2
 
 /// Usually `c3` has a continuation initialized by the whole code of the contract. It is used for function calls.
 /// The primitive returns the current value of `c3`.
--- a/tolk-tester/tests/cells-slices.tolk
+++ b/tolk-tester/tests/cells-slices.tolk
@@ -1,3 +1,5 @@
+import "@stdlib/exotic-cells"
+
 fun builder.store_u32(mutate self, value: int): self {
     return self.storeUint(value, 32);
 }
@@ -534,6 +536,30 @@ fun test41() {
     return data;
 }
 
+@method_id(142)
+fun test42() {
+    var c1 = beginCell().endCell();
+    var c2 = beginCell().endCellSpecial(false);
+    return (
+        c1.hash() == c1.hashOfLevel(3), 
+        c1.depth() == c1.depthOfLevel(3),
+        c1.hash() == c2.hash(),
+        c2.depth() == 0,
+    );
+}
+
+@method_id(143)
+fun test43() {
+    var c3 = beginCell().storeUint(123, 32).endCell();
+    var c3lib = c3.toLibraryReference();
+    
+    var (s, isExotic) = c3lib.beginParseSpecial();
+    assert (isExotic) throw 123;
+    val cellType = s.loadInt(8) as ExoticCellType;
+    return (cellType, s.loadUint(256) == c3.hash(), s.remainingBitsCount());
+    // 73669
+}
+
 fun main(): int {
     return 0;
 }
@@ -581,6 +607,8 @@ fun main(): int {
 @testcase | 138 |     | 0
 @testcase | 140 |     | 0 0
 @testcase | 141 |     | [ 9 9 9 9 9 9 9 9 9 8 8 8 8 2 ]
+@testcase | 142 |     | -1 -1 -1 -1
+@testcase | 143 |     | 2 -1 0
 
 We test that consequtive storeInt/storeUint with constants are joined into a single number
 
--- a/tolk/tolk-version.h
+++ b/tolk/tolk-version.h
@@ -18,6 +18,6 @@
 
 namespace tolk {
 
-constexpr const char* TOLK_VERSION = "1.1.0";
+constexpr const char* TOLK_VERSION = "1.2.0";
 
 } // namespace tolk


ProgramCrafter:
> Also, void is now used instead of never to indicate fields in structs that make no sense and can be missed out. Both these types are empty and mean "absence" or "non-existence", but never conflicts with control flow edges. Fixes https://github.com/ton-blockchain/ton/issues/1864 — an interesting corner case.

If I may, I would clarify the terminology used, regarding the word "empty".

The Tolk's type `void` (Rust: `()`, Idris2: `Unit`, C: `void`) has meaning "the value is already complete, no additional information required to create it".

The Tolk's type `never` (Rust: `!`, Idris2: `Void`, C does not have such type but only `[[ noreturn ]]` annotation) means "there cannot be a value of such type, no possible data is valid to create it", and having a variable of type `never` implies that control flow must exit the function somewhere before it. They typically call it the bottom type.

20787428717:
585807

alireza237237:
585808

PR #1924, opened 2025-11-28
Add `reverse-continue-to-actor-caller` GDB command

DanShaders:
This obviously only works under rr.

Debug flow looks like this:
```
Thread 2 received signal SIGABRT, Aborted.
(rr) where
...
#6  0x00005646dea4a2aa in ton::adnl::AdnlPeerTableImpl::subscribe
<frame from actor runtime>
#8  0x00005646dea1e82b in td::actor::core::ActorExecutor::flush_one_message
<other useless frames from actor scheduler; note that the caller location seems to be completely lost>
(rr) reverse-continue-to-actor-caller
(rr) where
<actor runtime internals>
#8  0x00005646de992953 in send_closure<td::actor::ActorId<ton::adnl::Adnl>&, <...>)
#9  ton::overlay::OverlayManager::register_overlay
<synchronous callers of `register_overlay`, frames from actor scheduler>
```
and honestly feels like magic if you ever tried to figure out actor caller manually.

### DIFF
--- a/.gdbinit
+++ b/.gdbinit
@@ -1,0 +1,1 @@
+source assembly/gdb/rc_to_actor_caller.py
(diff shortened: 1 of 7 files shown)


DanShaders:
> That version has not been published on PyPI

And I just invented a version to put to pyproject.toml, right?

> However, CMakeLists.txt:69-76 leaves TON_INSERT_GDB_HOOKS disabled by default

Works as intended. This is an advanced optional feature.

PR #1938, opened 2025-12-04
Clean secp256k1 build artifacts before build

awesome-doge:
## Summary

This PR fixes CI build failures by cleaning the `third-party/secp256k1` directory before starting the build process.

- Adds a cleanup step to remove stale autotools artifacts from `secp256k1` submodule
- Uses `git clean -fdx` and `git checkout .` to ensure a clean build state
- Prevents build failures caused by leftover files from previous builds

## Problem

The CI build occasionally fails due to stale autotools-generated files in the `secp256k1` submodule that conflict with fresh builds.

## Solution

Added a new step "Clean third-party build artifacts" that runs before the build step to:
1. Navigate to `third-party/secp256k1`
2. Remove all untracked files with `git clean -fdx`
3. Reset any modified files with `git checkout .`

### DIFF
--- a/.github/workflows/build-ton-linux-x86-64-shared.yml
+++ b/.github/workflows/build-ton-linux-x86-64-shared.yml
@@ -45,8 +45,8 @@ jobs:
       uses: actions/cache@v4
       with:
         path: ~/.ccache
-        key: ${{ runner.os }}-${{ runner.arch }}-${{ matrix.os }}-shared-ccache-${{ steps.date-stamp.outputs.timestamp }}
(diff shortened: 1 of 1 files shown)


DanShaders:
> The CI build occasionally fails due to stale autotools-generated files in the secp256k1 submodule that conflict with fresh builds.

Can you link an example workflow? I never seen this particular flake.

> `Invalidate ccache to fix illegal instruction error`

Thanks for debugging this! Can you please drop the commit though (I want to do a more comprehensive fix)?

PR #1939, opened 2025-12-04
Do not use -march=native in CI

DanShaders:
GitHub-provided CI runners are heterogeneous, so artifacts compiled with -march=native on a newer runner may fail to run on an older runner if ccache reuses same object files. To combat this, we introduce (and use in CI) TON_USE_CI_BASELINE_CPU CMake flag that overrides -march=native with some (relatively old) baseline CPU arch.


DanShaders:
This requires passing -march=${TON_ARCH} carefully to _all_ dependencies as well, which is extraordinarily annoying.

PR #1943, opened 2025-12-11
old: Add QUIC protocol implementation

avevad:


quic/quic-connection.h
```diff
@@ -0,0 +1,49 @@
...
+                                                              td::CSlice alpn = "ton");
+
+ protected:
+  void start_up() override;
+  void tear_down() override;
+  void hangup() override;
+  void hangup_shared() override;
+  void wake_up() override;
+  void alarm() override;
+  void loop() override;
```
DanShaders: You don't have to implement these if you don't use them. The default implementation is reasonable enough.

quic/quic-pimpl.cpp
DanShaders, line 1: I would define this in an anonymous namespace inside  `quic-connection.cpp` unless there is a legitimate reason why we want to import the implementation from multiple places.

quic/quic-pimpl.cpp
```diff
@@ -0,0 +1,221 @@
...
+  if (!ssl)
+    return td::Status::Error("SSL_new failed");
+
+  SSL_set_connect_state(ssl);
+
+  SSL_set_tlsext_host_name(ssl, host.c_str());
+
+  std::string alpn_data(alpn.size() + 1, '\0');
+  alpn_data[0] = static_cast<int8_t>(alpn.size());
+  std::copy_n(alpn.c_str(), alpn.size(), alpn_data.begin());
```
DanShaders: `alpn_data.begin() + 1` as the last argument?

quic/quic-pimpl.cpp
```diff
@@ -0,0 +1,221 @@
...
+  ngtcp2_settings_default(&settings);
+  settings.initial_ts = now_ts();
+
+  ngtcp2_transport_params params;
+  ngtcp2_transport_params_default(&params);
+
+  params.initial_max_streams_bidi = 4;
+  params.initial_max_stream_data_bidi_local = 1 << 20;
+  params.initial_max_stream_data_bidi_remote = 1 << 20;
+  params.initial_max_data = 1 << 20;
```
DanShaders: Where is this 1MB(?) coming from?

quic/quic-pimpl.cpp
```diff
@@ -0,0 +1,221 @@
...
+  dcid.datalen = 8;
+  scid.datalen = 8;
+  RAND_bytes(dcid.data, static_cast<int>(dcid.datalen));
+  RAND_bytes(scid.data, static_cast<int>(scid.datalen));
+
+  ngtcp2_path path{};
+  path.local.addr = const_cast<ngtcp2_sockaddr*>(local_address.get_sockaddr());
+  path.local.addrlen = static_cast<ngtcp2_socklen>(local_address.get_sockaddr_len());
+  path.remote.addr = const_cast<ngtcp2_sockaddr*>(remote_address.get_sockaddr());
+  path.remote.addrlen = static_cast<ngtcp2_socklen>(remote_address.get_sockaddr_len());
```
DanShaders, line 166: I don't like const casts here, is ngtcp2 guaranteed not to change passed `sockaddr`s?

quic/quic-pimpl.h
```diff
@@ -0,0 +1,89 @@
...
+    };
+
+    virtual void on_handshake_completed(const HandshakeCompletedEvent& event) = 0;
+    virtual void on_stream_data(const StreamDataEvent& event) = 0;
+
+    virtual ~Callback() = default;
+  };
+
+  constexpr static size_t QUIC_MTU = 1350;
+  constexpr static size_t UDP_MTU = 2048;
```
DanShaders: Why not proper 65507 bytes if this is already larger than ethernet MTU?

quic/quic-pimpl.h
```diff
@@ -0,0 +1,89 @@
...
+      bool fin = false;
+    };
+
+    virtual void on_handshake_completed(const HandshakeCompletedEvent& event) = 0;
+    virtual void on_stream_data(const StreamDataEvent& event) = 0;
+
+    virtual ~Callback() = default;
+  };
+
+  constexpr static size_t QUIC_MTU = 1350;
```
DanShaders: Where does this come from? I think ngtcp2 does some form of dynamic MTU discovery, so we should probably query the value from it or use true maximum

quic/quic-pimpl.cpp
```diff
@@ -0,0 +1,221 @@
...
+  if (ngtcp2_crypto_ossl_configure_client_session(ssl) != 0)
+    return td::Status::Error("ngtcp2_crypto_ossl_configure_client_session failed");
+
+  if (ngtcp2_crypto_ossl_ctx_new(&ossl_ctx, ssl) != 0)
+    return td::Status::Error("ngtcp2_crypto_ossl_ctx_new failed");
+
+  return td::Status::OK();
+}
+
+td::Status QuicConnectionPImpl::init_quic() {
```
DanShaders: Why is this not a constructor?

CMakeLists.txt
```diff
@@ -186,6 +186,14 @@ if (USE_LIBRAPTORQ)
   add_subdirectory(third-party/libraptorq EXCLUDE_FROM_ALL)
 endif()
 
+option(USE_QUIC "use ngtcp2 QUIC library" OFF)
+if (USE_QUIC)
+  find_package(OpenSSL 3.2 REQUIRED COMPONENTS SSL Crypto)
```
DanShaders: We build with OpenSSL 3.1.4 in CI. Not sure how this passed, I think this might just never execute as it sees previously set OpenSSL_FOUND (or something).

quic/quic-pimpl.cpp
avevad: bc the work it does is too complex to reside in a constructor + we need to return td::Status too

quic/quic-connection.h
avevad: i'm not really familiar with td actor system, so i left them for debug logs in case of some unexpected signal from the scheduler

quic/quic-pimpl.cpp
avevad: that means you will need to declare all its ngtcp-insides in quic-connection.h, too
this is highly unwanted, as ngtcp exports some weird C symbols into global namespace -- they will contaminate the scope of every TU it's included into
see [PImpl idiom](https://en.cppreference.com/w/cpp/language/pimpl.html)

DanShaders: Thanks, I know what pimpl is. I mostly don't like a separate header for implementation when this can be done in a single file, i. e. something resembling this: https://github.com/DanShaders/ton/blob/alpenglow/validator/consensus/block-producer.cpp#L11-L13.

quic/quic-pimpl.cpp
avevad: fixed

quic/quic-pimpl.cpp
avevad: anyways, i plan to reuse this impl structure in server-side connections, so it will be useful to keep it in separate files

PR #1989, opened 2025-12-27
Remove spaces at EOL

abacabadabacaba:
Remove whitespace at EOL and add a lint to prevent it from reappearing.

### DIFF
--- a/.github/workflows/build-ton-linux-x86-64-werror.yml
+++ b/.github/workflows/build-ton-linux-x86-64-werror.yml
@@ -42,6 +42,10 @@ jobs:
           -DTON_WERROR_BUILD=On
         ninja -C build
 
+    - name: Check spaces at EOL
+      run: |
+        ! git grep --cached -Il '\s$' :!example/android/third_party
+
     - name: Check C++ formatting
       run: |
         # TODO: Run clang-format only on changed files.


example/android/third_party/libsodium/libsodium-android-armv7-a/lib/pkgconfig/libsodium.pc
DanShaders, line 1: It's probably a good idea to exclude example/android/third_party

.github/workflows/build-ton-linux-x86-64-werror.yml
```diff
@@ -42,6 +42,11 @@ jobs:
           -DTON_WERROR_BUILD=On
         ninja -C build
 
+    - name: Check spaces at EOL
+      run: |
+        find -path ./.git -prune -o -path ./third-party -prune -o -path ./LICENSE.LGPL -o -type f ! -name \*.a ! -name \*.so ! -name \*.jar ! -name \*.boc ! -name \*.png -exec sed -i 's/\s\+$//' {} +
```
DanShaders: Maybe `sed -i -e 's/[ \t]\+\r$/\r/' -e 's/[ \t]\+$//' $(git grep -Il '' -- ':!^example/android/third_party')`?
This automatically excludes binary files and files in .gitignore and doesn't turn CRLF to LF (which then makes Git to output warning in CI)

example/android/third_party/libsodium/libsodium-android-armv7-a/lib/pkgconfig/libsodium.pc
abacabadabacaba: Done.

.github/workflows/build-ton-linux-x86-64-werror.yml
abacabadabacaba: I made an even better version.

PR #1992, opened 2025-12-31
Repair packages after renaming files in archive

SpyCheese:


Edgar71769:
> **Issues**
> 
> * validator/import-db-slice-local.cpp:71-139: `Package::iterate` now returns `Status`, but the call ends with `.ignore()`. If a package is corrupt (too short/broken), iteration errors are silently discarded and `process_package` returns OK with nothing loaded, masking import failures. Capture and propagate/log the status instead.
> * validator/import-db-slice.cpp:115-162: Same pattern—iteration status is ignored. Corrupt archives will now be silently skipped, making import failures hard to diagnose. Handle the returned `Status`.
> * validator/manager.cpp:1790-1821: `iterate()` errors are ignored when scanning import packages for proofs, so malformed packages now fail silently and can trigger confusing “proof not found” results. Check the status and log/abort accordingly.

C

Edgar71769 (commented):
.

PR #2097, opened 2026-01-26
Add OpenMetrics exporter (with QUIC stats)

avevad:


EmelyanenkoK:
@codex please review

quic/quic-common.h
```diff
@@ -14,6 +14,31 @@
...
+  int64_t bytes_unacked = 0, bytes_unsent = 0;
+  int64_t total_sids = 0, open_sids = 0;
+  double mean_rtt = 0;
+
+  QuicConnectionStats operator+(const QuicConnectionStats &other) const {
+    return {
+      .bytes_rx = bytes_rx + other.bytes_rx, .bytes_tx = bytes_tx + other.bytes_tx,
+      .bytes_lost = bytes_lost + other.bytes_lost, .bytes_unacked = bytes_unacked + other.bytes_unacked,
+      .bytes_unsent = bytes_unsent + other.bytes_unsent, .total_sids = total_sids + other.total_sids,
+      .open_sids = open_sids + other.open_sids
```
DanShaders: If you add comma on the last like, clang-format will do a much better job.

quic/quic-common.h
```diff
@@ -14,6 +14,31 @@
...
+    };
+  }
+
+  QuicConnectionStats operator-(const QuicConnectionStats &other) const {
+    return {
+      .bytes_rx = bytes_rx - other.bytes_rx, .bytes_tx = bytes_tx - other.bytes_tx,
+      .bytes_lost = bytes_lost - other.bytes_lost, .bytes_unacked = bytes_unacked - other.bytes_unacked,
+      .bytes_unsent = bytes_unsent - other.bytes_unsent, .total_sids = total_sids - other.total_sids,
+      .open_sids = open_sids - other.open_sids
```
DanShaders: Similarly here

quic/quic-pimpl.cpp
```diff
@@ -524,6 +524,22 @@ td::Status QuicConnectionPImpl::buffer_stream(QuicStreamID sid, td::BufferSlice
...
+  size_t bytes_unacked = 0, bytes_unsent = 0;
+  for (auto &[_, stream] : streams_) {
+    bytes_unacked += stream.pin_.size();
+    bytes_unsent += stream.reader_.size();
+  }
+  return {
+    .bytes_rx = static_cast<int64_t>(info.bytes_recv), .bytes_tx = static_cast<int64_t>(info.bytes_sent),
+    .bytes_lost = static_cast<int64_t>(info.bytes_lost), .bytes_unacked = static_cast<int64_t>(bytes_unacked),
+    .bytes_unsent = static_cast<int64_t>(bytes_unsent), .total_sids = static_cast<int64_t>(sids_encountered),
+    .open_sids = static_cast<int64_t>(streams_.size()), .mean_rtt = static_cast<double>(info.smoothed_rtt)
```
DanShaders: And here :)

test/net/bench-rldp.cpp
```diff
@@ -486,12 +495,18 @@ void run_client(Config config) {
...
   td::actor::ActorOwn<BenchmarkRunner> runner;
+  td::actor::ActorOwn<ton::PrometheusExporter> exporter;
   td::actor::ActorOwn<StatsReporter> stats_reporter;
 
   ton::adnl::AdnlNodeIdShort src;
   ton::adnl::AdnlNodeIdShort dst;
 
   scheduler.run_in_context([&] {
+    if (config.prometheus) {
+      exporter = ton::PrometheusExporter::listen(29777);
```
DanShaders: Can we please not add more servers that are not bound to any IP? We have enough already and any such instance leads to `# FIXME` comment in tontester. Also, please make ip + port configurable with a cmdline option.

quic/quic-common.h
avevad: thanks!

test/net/bench-rldp.cpp
avevad: fixed

metrics/metrics-collectors.h
```diff
@@ -0,0 +1,270 @@
...
+  virtual void collect(MetricsPromise P) = 0;
+  virtual ~AsyncCollector() = default;
+};
+
+using AsyncCollectorClosure = std::function<void(MetricsPromise)>;
+
+// Actors with simple metrics collection should virtually inherit this instead.
+// Also, they **MUST** trivially override `collect` function, because send_closure implementation is not really handy
+// (otherwise `&A::collect` will have type `void (CollectorWrapper::*)(MetricsPromise)`, which is unrelated to
+//  `td::actor::Actor` -- so `send_closure`'s inference of member function class will fail).
```
DanShaders, line 36: That's cringe. Use send_lambda in the collector to call the method

metrics/metrics-collectors.h
```diff
@@ -0,0 +1,270 @@
...
+
+ private:
+  const std::string label_name_;
+  std::function<std::shared_ptr<InstrumentType>()> make_;
+  std::unordered_map<LabelType, std::shared_ptr<InstrumentType>> instruments_;
+  std::mutex mutex_;
+};
+
+template <typename A>
+void CollectorWrapper::add_collector(td::actor::ActorId<A> collector) {
```
DanShaders, line 164: To the end of file: is there any reason these are defined out-of-line? Seems like it just makes it harder to read

validator-engine/validator-engine.hpp
```diff
@@ -423,6 +425,7 @@ class ValidatorEngine : public td::actor::Actor {
 
   void start_up() override;
   ValidatorEngine() {
+    exporter_ = ton::PrometheusExporter::create();
```
DanShaders, line 428: Why does (a) PrometheusExporter require a factory method instead of constructor with defaulted params (b) does not get created somewhere after `start_up` as do all other actors?

quic/quic-server.h
```diff
@@ -87,6 +87,31 @@ class QuicServer : public td::actor::Actor, public td::ObserverBase {
...
+        return res;
+      }
+
+      Entry operator-(const Entry &other) const {
+        Entry res = {.total_conns = total_conns - other.total_conns, .impl_stats = impl_stats - other.impl_stats};
+        res.impl_stats.mean_rtt = impl_stats.mean_rtt;
+        return res;
+      }
+
+    } summary = {};
```
DanShaders, line 109: I am not amused by struct, field, and deffault value declaration in one statement

quic/quic-server.h
```diff
@@ -87,6 +87,31 @@ class QuicServer : public td::actor::Actor, public td::ObserverBase {
                                                             std::unique_ptr<Callback> callback, td::Slice alpn,
                                                             td::Slice bind_host, Options options);
 
+  struct Stats {
+    struct Entry {
+      size_t total_conns = 1;
```
DanShaders, line 92: So total_conns cannot be less than 1?

metrics/metrics-collectors.cpp
```diff
@@ -0,0 +1,75 @@
...
+}
+
+td::actor::Task<MetricSet> CollectorWrapper::collect_coro() {
+  MetricSet whole_set = {};
+  for (auto &f : collector_closures_) {
+    auto [future, promise] = td::actor::StartedTask<metrics::MetricSet>::make_bridge();
+    f(std::move(promise));
+    auto metric_set = co_await std::move(future);
+    whole_set = std::move(whole_set).join(std::move(metric_set));
+  }
```
DanShaders, line 18: So this will wait each collector sequentially? We should probably start all collectors simultaneously and use `td::actor::all` / start all loop + wait loop

http/http-server.cpp
```diff
@@ -23,6 +23,15 @@ namespace ton {
 
 namespace http {
 
+HttpServer::HttpServer(td::IPAddress address, std::shared_ptr<Callback> callback)
+    : address_(address), callback_(std::move(callback)) {
+  add_collector(collector_.get());
+  td::actor::send_closure(collector_.get(), &metrics::MultiCollector::add_sync_collector, metrics_.connections);
+  td::actor::send_closure(collector_.get(), &metrics::MultiCollector::add_sync_collector, metrics_.connections_total);
+  td::actor::send_closure(collector_.get(), &metrics::MultiCollector::add_sync_collector, metrics_.requests_total);
+  td::actor::send_closure(collector_.get(), &metrics::MultiCollector::add_sync_collector, metrics_.responses_total);
```
DanShaders, line 32: This is at least somewhat weird to see, we usually don't use actor methods from inside constructor

metrics/metrics-example.cpp
```diff
@@ -0,0 +1,47 @@
...
+  void collect(metrics::MetricsPromise P) override;
+
+ private:
+  metrics::MultiCollector::Own collector_ = metrics::MultiCollector::create("example");
+  metrics::LambdaCounter::Ptr time_counter_ = metrics::LambdaCounter::make(
+      "current_time_seconds",
+      [] { return std::vector{metrics::Sample{.label_set = {}, .value = td::Timestamp::now().at_unix()}}; },
+      "Number of seconds passed since January 1, 1970");
+  metrics::LambdaGauge::Ptr stack_gauge_ = metrics::LambdaGauge::make("current_stack_top_bytes", [] {
+    void* stack_top = &stack_top;
```
DanShaders, line 22: wtf
I mean I see how this works but this is probably ub for 20 reasons at once

metrics/metrics-types.cpp
```diff
@@ -0,0 +1,155 @@
+#include <unordered_map>
+
+#include "td/utils/logging.h"
+
+#include "metrics-types.h"
+
+namespace ton::metrics {
+
+std::string concat_names(std::string name1, std::string name2) {
```
DanShaders, line 9: Should be static / in anonymous namespace

metrics/metrics-collectors.h
```diff
@@ -0,0 +1,270 @@
...
+  virtual ~AsyncCollector() = default;
+};
+
+using AsyncCollectorClosure = std::function<void(MetricsPromise)>;
+
+// Actors with simple metrics collection should virtually inherit this instead.
+// Also, they **MUST** trivially override `collect` function, because send_closure implementation is not really handy
+// (otherwise `&A::collect` will have type `void (CollectorWrapper::*)(MetricsPromise)`, which is unrelated to
+//  `td::actor::Actor` -- so `send_closure`'s inference of member function class will fail).
+class CollectorWrapper : public AsyncCollector {
```
DanShaders, line 37: I cannot get rid of the thought that this class separation is overengineered.
What are the use cases for implementing `AsyncCollector` but not `CollectorWrapper`? Why would one implement `CollectorWrapper` without delegating to `MultiCollector`? Why `MultiCollector` needs to be a separate actor as we can just always implement `AsyncCollector` on the actor that actually has metrics?

metrics/metrics-example.cpp
DanShaders: Ah there is `__builtin_frame_address(0)`, use that

PR #2098, opened 2026-01-27
OpenSSL 3.5

neodix42:
- OpenSSL upgrade to 3.5;
- OpenSSL, libsodium, lz4  moved to submodules;
- OpenSSL linked statically to TON binaries where required;
- reworked Android tonlib assembly (removed prebuilt static third-party libs);
- added msys2 mingw64 and ucrt64 build scripts.

### DIFF
--- a/test/tontester/src/consensus_explorer/explorer.py
+++ b/test/tontester/src/consensus_explorer/explorer.py
@@ -1,7 +1,6 @@
 from multiprocessing import Process
 from pathlib import Path
-from typing import final, cast
-import os
+from typing import cast, final
 
 from .parser import ParserLogs
 from .visualizer import DashApp
@@ -40,9 +39,15 @@ def _main():
     import argparse
 
     parser = argparse.ArgumentParser()
-    _ = parser.add_argument("--logs", nargs="+", required=True, help="Paths to log files or directory")
-    _ = parser.add_argument("--host", default="127.0.0.1", help="Host to bind to (default: 127.0.0.1)")
-    _ = parser.add_argument("--port", type=int, default=8050, help="Port to bind to (default: 8050)")
+    _ = parser.add_argument(
+        "--logs", nargs="+", required=True, help="Paths to log files or directory"
+    )
+    _ = parser.add_argument(
+        "--host", default="127.0.0.1", help="Host to bind to (default: 127.0.0.1)"
+    )
+    _ = parser.add_argument(
+        "--port", type=int, default=8050, help="Port to bind to (default: 8050)"
+    )
 
     args = parser.parse_args()
     logs = cast(list[str], args.logs)
--- a/test/tontester/src/consensus_explorer/parser/parser_base.py
+++ b/test/tontester/src/consensus_explorer/parser/parser_base.py
@@ -1,4 +1,5 @@
 from abc import ABC, abstractmethod
+
 from ..models import ConsensusData
 
 
--- a/test/tontester/src/consensus_explorer/parser/parser_logs.py
+++ b/test/tontester/src/consensus_explorer/parser/parser_logs.py
@@ -4,8 +4,8 @@ from datetime import datetime
 from pathlib import Path
 from typing import final, override
 
+from ..models import ConsensusData, EventData, SlotData
 from .parser_base import Parser
-from ..models import ConsensusData, SlotData, EventData
 
 type slot_id_type = tuple[str, int]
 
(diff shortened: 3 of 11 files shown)


tdutils/td/utils/AtomicRead.h
DanShaders, line 1: I don't think this is correct. Let's stop pretending we are smarter than stdlib authors and use std::mutex instead of this.

neodix42: Agree, not the obvious fix either. Will be rolled back in this PR.

neodix42:
> As per discussion in Telegram, let's split Windows code patches from this PR. In particular, I don't want to see any MSVC workarounds in validator/consensus. If we need a native Windows build of validator so much, let's use clang-cl.
> 
> I am not a big fan of msys2 mingw and ucrt builds but this will be a concern for a later "more Windows support" PR.
> 
> I'll let @birydrad or @avevad comment on how much they like or dislike platform ifdefs in QUIC (I personally think they are incredibly ugly but that's not my code).

I don’t want to split Windows patches, since all code must be able to compile against MSVC — we have external projects relying on this.
Good point regarding clang-cl though; this may come in the future as a separate PR.
MinGW and UCRT builds have proven to be stable and show reasonable potential, so I’d prefer to leave them for now.

.github/workflows/build-ton-linux-android-tonlib.yml
```diff
@@ -8,16 +8,24 @@ jobs:
 
     steps:
     - name: Check out repository
-      uses: actions/checkout@v3
+      uses: actions/checkout@v4
```
Danil42Russia, line 11: Why not `v6`? All new changes will only be available for `6.x.y` versions

.github/workflows/build-ton-wasm-emscripten.yml
```diff
@@ -24,13 +24,6 @@ jobs:
         chmod +x llvm.sh
         sudo ./llvm.sh 21 clang
 
-    - name: Cache OpenSSL
```
Danil42Russia, line 27: I may be wrong, but this should also be removed from:
- `build-ton-linux-arm64-appimage.yml`
- `build-ton-linux-arm64-shared.yml`
- `build-ton-linux-x86-64-shared.yml`

.github/workflows/build-ton-linux-x86-64-appimage.yml
Danil42Russia, line 67: https://github.com/neodix42/ton/blob/7d1995ec0a03bac280bdf63c91f005696d3ab816/assembly/appimage/create-appimages.sh#L30

```
cp ../openssl_3/libcrypto.so.3 \
```

So the `openssl_3` folder is gone, which will cause problems, won't it?

crypto/CMakeLists.txt
```diff
@@ -292,10 +292,18 @@ if (NOT WIN32)
   target_link_libraries(ton_crypto_core PUBLIC dl z)
 endif()
 target_include_directories(ton_crypto_core SYSTEM PUBLIC $<BUILD_INTERFACE:${OPENSSL_INCLUDE_DIR}>)
+# Ensure OpenSSL is built before ton_crypto_core
```
DanShaders, line 295: .. and everywhere else where you add this dependency:

When building OpenSSL manually, provide target `OpenSSL::Crypto` (just like `find_package(OpenSSL)`) and then add it instead of `${OPENSSL_CRYPTO_LIBRARY}` to `target_link_libraries`. This way everything linking to OpenSSL will have it built.

.github/workflows/build-ton-wasm-emscripten.yml
neodix42: No, we need all of them. Shared ones for simple compilation test of these platforms, and the appimage generates artifacts for arm64 linux platforms.

.github/workflows/build-ton-linux-x86-64-appimage.yml
neodix42: it wo't fail, since the [script](https://github.com/neodix42/ton/blob/7d1995ec0a03bac280bdf63c91f005696d3ab816/assembly/native/build-ubuntu-appimages.sh)  still uses openssl 3 in old fashion. I will rework it, good point.

.github/workflows/build-ton-linux-android-tonlib.yml
neodix42: I will test with v6 and update if no issues.

crypto/CMakeLists.txt
neodix42: I thought about it, but we have multiple targets and not on all of them it worked, so I went for a most generic approach.

PR #2114, opened 2026-02-01
move zlib and microhttpd into submodules

neodix42:


.gitignore
```diff
@@ -17,11 +17,8 @@ test/regression-tests.cache/
...
 .DS_Store
 dev/
-zlib/
-libsodium/
 libmicrohttpd-0.9.77-w32-bin/
 readline-5.0-1-lib/
-openssl-3.5/
 libsodium-1.0.18-stable-msvc.zip
 libmicrohttpd-0.9.77-w32-bin.zip
 openssl-3.5.zip
```
Danil42Russia: now this can be removed, after all, in git modules

CMake/BuildMHD.cmake
```diff
@@ -0,0 +1,160 @@
...
+        RANLIB=${TON_ANDROID_RANLIB}
+        CFLAGS=${MHD_CFLAGS}
+        make clean
+      COMMAND ${CMAKE_COMMAND} -E env
+        CC=${TON_ANDROID_CC}
+        CXX=${TON_ANDROID_CXX}
+        AR=${TON_ANDROID_AR}
+        RANLIB=${TON_ANDROID_RANLIB}
+        CFLAGS=${MHD_CFLAGS}
+        make -j16
```
Danil42Russia, line 99: There may be an unexpected effect when specifying `-j` less than `16` for a project (and even more so when there are fewer than 16 cores on the system) :(

I may be mistaken, but if you simply specify `-j` without the number of cores, the value from the project will be implicitly set. I need to check this; I'm not sure if it will work without the `ExternalProject` module.
see:
1) https://www.gnu.org/software/make/manual/html_node/Options_002fRecursion.html
2) https://cmake.org/cmake/help/latest/module/ExternalProject.html

PR #2196, opened 2026-03-11
Disable QUIC connection flood control in tontester

DanShaders:
When we run a local network, we (unsurprisingly) have a million connections from localhost.


EmelyanenkoK (approved):

care44cubs-max (commented):
Yes

PR #2202, opened 2026-03-13
Multiple QUIC fixes

avevad:
Fix RPK parse failure leading to abort

Fix stream limits

Fix stream MTU behavior

Fix OOB in quic cid random

Fix OOB in CC algo selection


EmelyanenkoK (approved):

care44cubs-max (commented):
Yes

PR #2211, opened 2026-03-17
Fix empty collated data

Mustang98:
- Fixed behavior mismatch between compression and decompression of empty collated data. 
- Improved external interface of boc.deserialize() method. Now it returns an error status instead of a negative size in case of header parsing error. It simplifies result processing in caller sites and avoids confusion.


kingdecember4-tech:
?

SpyCheese (approved):

care44cubs-max (commented):
Yes

PR #2296, opened 2026-04-10
Add checks for validators weight overflow

EmelyanenkoK:
Although these checks are not important in practice because validator weight is limited in `elector-code.fc`, the possibility of weight overflow has been reported to the bug bounty program dozens of times. To avoid wasting security researchers’ time, we decided to add these checks explicitly.


Robyn37 (commented):
128a711941418b5fe499f0799e2298799740df28

EmelyanenkoK:
Closed in favor of https://github.com/ton-blockchain/ton/pull/2343

PR #2298, opened 2026-04-12
Add an environment variable to disable stack traces

abacabadabacaba:


DanShaders:
Why not just compile with `TON_DISABLE_BACKTRACE`?

abacabadabacaba:
I think that it should be possible to disable debugging features like stacktraces even after TON is built/installed. But maybe it's not really needed.

DanShaders:
Superseded by #2356

PR #2335, opened 2026-04-27
Make BitArray a little constexpr-friendly

abacabadabacaba:


DanShaders (approved):
Removes `operator<` for `n % 8 != 0` case and the explicit defaulted copy constructor. Don't think this matters much though.

PR #2353, opened 2026-05-07
Add support for prehandlers in bus runtime

DanShaders:


DanShaders:
Will do this a different way

Issue #2351, opened 2026-05-07
validator-engine has memory leak in the server


nooperpudd:
command
```
validator-engine -c /var/ton-work/db/config.json -C /var/ton-work/db/ton-global.config --db /var/ton-work/db --state-ttl 2592000 --archive-ttl 2592000 --threads 15 --verbosity 1 --celldb-cache-size=64424509440 --celldb-preload-all --celldb-direct-io --fast-state-serializer
```

<img width="710" height="299" alt="Image" src="https://github.com/user-attachments/assets/b6ff2955-91b3-4289-871b-3d28f5388ce9" />

DanShaders:
Yes. We are planning to address one of known leaks in the next update.

PR #2355, opened 2026-05-10
prevent CellDb preload from inflating RocksDB cache

SIDDHANTCOOKIE:
@DanShaders would love your thoughts on this 
Add iteration options to the key-value reader path and use RocksDB `ReadOptions.fill_cache = false` for CellDb preload.
Mitigates #2351 
### What changed

- Added `KeyValueReader::ForEachOptions`
- Added `for_each(..., options)` support in:
  - `RocksDb`
  - `MemoryKeyValue`
- Updated CellDb preload to call `for_each(..., {.fill_cache = false})`

### Effect

- `--celldb-preload-all` still scans the whole DB
- preload no longer aggressively fills RocksDB block cache
- should reduce startup / steady-state RAM inflation in large-cache validator setups

### DIFF
--- a/tddb/td/db/KeyValue.h
+++ b/tddb/td/db/KeyValue.h
@@ -60,11 +60,20 @@ class KeyValueReader {
  public:
   virtual ~KeyValueReader() = default;
   enum class GetStatus : int32 { Ok, NotFound };
+  struct ForEachOptions {
+    bool fill_cache = true;
+  };
 
   virtual Result<GetStatus> get(Slice key, std::string& value) = 0;
   virtual Result<std::vector<GetStatus>> get_multi(td::Span<Slice> keys, std::vector<std::string>* values) = 0;
   virtual Result<size_t> count(Slice prefix) = 0;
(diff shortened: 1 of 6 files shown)


DanShaders:
I'm not too familiar with the db code but isn't the whole point of preload to populate caches? I also thought #2351 is about runtime leakage (which we think happens because we leak rocksdb snapshots), not start up memory consumption

SIDDHANTCOOKIE:
Hmm good point, this patch is about preload-driven cache growth, not runtime snapshot leakage. After looking through the code, I agree it probably doesn’t address #2351 as stated. I’ll investigate snapshot lifetime / old loader retention directly, especially using the existing `td.rocksdb.snapshot.oldest_snapshot_ago.seconds` stats, will get back on this

albertoblue87-netizen (commented):
Done

albertoblue87-netizen (commented):
Done

albertoblue87-netizen (commented):
Done

PR #2374, opened 2026-05-15
test/consensus: fix stop_instance loop using wrong node index

KOKosaaaa:
## What

Fix a copy-paste typo in the inner loop of `TestConsensus::finalize()` in `test/validator/consensus/test-consensus.cpp`.

```diff
   for (size_t idx = 0; idx < N_NODES; ++idx) {
-    for (size_t i = 0; i < nodes_[i].instances.size(); ++i) {
+    for (size_t i = 0; i < nodes_[idx].instances.size(); ++i) {
       tasks.push_back(stop_instance(idx, i));
     }
   }
```

The inner loop checked `nodes_[i].instances.size()` (using the inner counter `i` as the node index) instead of `nodes_[idx].instances.size()`. The parallel loop just below at lines 832-836 already uses the correct `nodes_[idx]` form.

## Why it matters

When instance counts are heterogeneous (e.g. `--n-double-nodes N > 0`), the inner `i` can advance past `nodes_[idx].instances.size()`. The subsequent `stop_instance(idx, i)` call then does:

```cpp
Instance &inst = node.instances[instance_idx];   // out-of-range
...
inst.bus.publish<StopRequested>();               // UB / SIGSEGV
```

## Reproducer (before this PR)

```bash
./test-consensus --duration 30 --n-nodes 5 --n-double-nodes 2                  --gremlin-period 5 --gremlin-kills-leader
```

Crashes with `Signal: 11` and a backtrace ending in `TestConsensus::stop_instance` / `BusHandle::publish<StopRequested>` during teardown. The crash is timing-sensitive (it needs at least two double-nodes plus a leader-killing gremlin to manifest reliably) because the bad iteration only triggers UB when scheduling lines up with concurrent stops.

## Verification

After the fix, the same command runs to completion and `TEST RESULTS` are printed for every node with no signal.

### DIFF
--- a/test/validator/consensus/test-consensus.cpp
+++ b/test/validator/consensus/test-consensus.cpp
@@ -824,7 +824,7 @@ class TestConsensus : public td::actor::Actor {
     LOG(WARNING) << "TEST FINISHED";
     std::vector<td::actor::Task<>> tasks;
     for (size_t idx = 0; idx < N_NODES; ++idx) {
-      for (size_t i = 0; i < nodes_[i].instances.size(); ++i) {
+      for (size_t i = 0; i < nodes_[idx].instances.size(); ++i) {
         tasks.push_back(stop_instance(idx, i));
       }
     }


KOKosaaaa:
Verified the fix locally by applying it to the test source, rebuilding `test-consensus`, and running the previously-crashing scenarios.

```
$ sed -i 's|nodes_\[i\].instances.size|nodes_[idx].instances.size|' test/validator/consensus/test-consensus.cpp
$ ninja test-consensus

# Before the fix: SIGSEGV reproducible.
# After the fix:

$ ./test-consensus --duration 30 --n-nodes 9 --n-double-nodes 2                    --gremlin-period 5 --gremlin-kills-leader -v 2
...
TEST RESULTS:
Node #0 instance #0 : synced up to block 14
Node #0 instance #1 : synced up to block 14
Node #1 instance #0 : synced up to block 14
Node #1 instance #1 : synced up to block 14
Node #2 instance #0 : synced up to block 14
Node #3 instance #0 : synced up to block 14
Node #4 instance #0 : synced up to block 14
Node #5 instance #0 : synced up to block 14
Node #6 instance #0 : synced up to block 14
Node #7 instance #0 : synced up to block 14
Node #8 instance #0 : synced up to block 14
Test finished
```

Also ran the 5-node repro 5 times back-to-back: 5/5 exit 0, no `Signal: 11`. Pre-patch the same command reproduced the crash with ~67% probability per run.

KOKosaaaa:
Noticed after submission that #2244 already proposes the same one-line fix and has been open since 2026-03-29. Apologies for the duplicate, I missed it in my pre-submission search.

Happy to close this PR if maintainers prefer to land #2244 first. The independent reproduction and verification details above may still be useful as a second data point for the same bug.

DanShaders:
Yes, let's close in favor of the previous PR

KOKosaaaa:
Closing as duplicate of #2244. The upstream PR proposes the same fix and was opened first.

PR #2409, opened 2026-05-30
VLOG categories

birydrad:
Adds per-category verbosity control on top of the existing global verbosity level.

- LogCategory with DECLARE_LOG_CATEGORY / DEFINE_LOG_CATEGORY and a variadic VLOG(category, LEVEL) << .... Each category inherits the global level until explicitly overridden, then tracks its own level.
- Migrated existing per-subsystem VLOG usages (adnl, dht, overlay, rldp/rldp2, validator, full-node, collator-node) to named categories and removed the old *_WARNING/*_INFO-style prefix constants in favour of
  plain INFO/WARNING/… levels per category.
- Configure at startup: validator-engine --vcategory adnl=5,overlay=2 (and --vcategory list to print categories).
- Adjust at runtime from the validator console.

### DIFF
--- a/tl/generate/scheme/ton_api.tl
+++ b/tl/generate/scheme/ton_api.tl
@@ -807,6 +807,8 @@ engine.validator.getStats = engine.validator.Stats;
 engine.validator.getConfig = engine.validator.JsonConfig;
 
 engine.validator.setVerbosity verbosity:int = engine.validator.Success;
+engine.validator.setLogCategoryVerbosity name:string verbosity:int = engine.validator.Success;
+engine.validator.getLogCategories = engine.validator.TextStats;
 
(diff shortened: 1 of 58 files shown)


tl/generate/scheme/ton_api.tl
```diff
@@ -807,6 +807,8 @@ engine.validator.getStats = engine.validator.Stats;
 engine.validator.getConfig = engine.validator.JsonConfig;
 
 engine.validator.setVerbosity verbosity:int = engine.validator.Success;
+engine.validator.setLogCategoryVerbosity name:string verbosity:int = engine.validator.Success;
+engine.validator.getLogCategories = engine.validator.TextStats;
```
DanShaders, line 811: Add these to `tonlib/EngineConsoleClient.cpp`.

SpyCheese (approved):

PR #2414, opened 2026-06-01
Clamp slots_per_leader_window to at least 1

kesha-antonov:
`slots_per_leader_window` (consensus config, param 30) is used directly as a divisor in the Simplex leader schedule, e.g. in `validator/consensus/simplex/bus.cpp`:

    slot / slots_per_leader_window_ % leaders_count_

and similarly in `consensus.cpp`, `pool.cpp` and `db.cpp`. It is read from the config in `crypto/block/mc-config.cpp` without any lower bound, so a value of 0 results in an integer division by zero and crashes the validator.

This clamps it to at least 1 when the config is parsed. Changing param 30 requires a config-change vote, so it is not reachable by an external party - it is just defense in depth, but a divisor coming straight from config seemed worth guarding against.


DanShaders:
See #2415

PR #2420, opened 2026-06-05
Remove handling of --local-config

DanShaders:
This is a lot of code for the feature that nobody uses.


samsuryajho-maker (commented):
9087#

PR #2429, opened 2026-06-08
Improve recovery sync over custom overlays

tvorogme:
## Summary

This PR makes custom overlays usable for validator recovery and catch-up, not only for live block broadcasts.

The main goal is to let operators with trusted/private archive nodes recover faster after restart while preserving the public overlay as the fallback path. A node can first try custom overlay peers for archive slices, block downloads, and next-block catch-up; if that path is unavailable or returns an error, the existing public overlay path remains available.

## What changed

- Custom overlays now answer archive and block sync queries:
  - `getArchiveInfo`, `getShardArchiveInfo`, `getArchiveSlice`
  - `downloadBlockFull`, `downloadNextBlockFull`
  - block proof/proof-link preparation and download queries
- Full node recovery now tries custom overlay peers for:
  - archive slice download
  - block download
  - next-block catch-up
  - block proof/proof-link downloads
- Custom archive recovery can resolve configured ADNL peers through DHT before downloading, so a custom overlay config with short ADNL ids can still be used for direct peer downloads.
- Public overlay fallback is kept for unsupported shards, unavailable custom actors, failed custom downloads, or missing peers.
- Shard client now buffers masterchain notifications that arrive while it is busy, then applies the exact next pending notification after saving state to DB. This avoids dropping live MC notifications during archive-to-live handoff.
- Archive recovery stays closer to live before handoff (`80s` freshness window reduced to `8s`) to reduce the post-archive live sync gap.

## Why

Before this change, custom overlays could carry live block broadcasts but could not serve archive recovery queries. A restarted node could see fresh private-overlay broadcasts but still had to fill historical gaps through public overlay peers. During the handoff from archive recovery to live sync, shard-client could also drop fresh masterchain notifications while busy and later rediscover them through the slower state-wait path.

With this PR, private/custom overlay peers can participate in the full restart recovery path, while public overlay behavior remains the safety net.

## Safety

- Custom overlay downloads only change peer selection and transport path.
- Downloaded archives, blocks, proofs, and states still go through the existing validation/import paths.
- Shard-client pending notification storage is bounded and only applies the exact next masterchain block id.
- Public overlay fallback remains available when custom sync is not possible.

## Validation

Built locally from `testnet` with:

```bash
SDKROOT=$(xcrun --show-sdk-path) LC_ALL=C LANG=C cmake -S . -B build-testnet-baseline -DCMAKE_BUILD_TYPE=Release -DTON_USE_JEMALLOC=OFF
SDKROOT=$(xcrun --show-sdk-path) LC_ALL=C LANG=C cmake --build build-testnet-baseline --target validator-engine validator -j6
```

Both `validator-engine` and `validator` built successfully before and after the patch.

### DIFF
--- a/validator-engine/validator-engine.hpp
+++ b/validator-engine/validator-engine.hpp
@@ -278,7 +278,7 @@ class ValidatorEngine : public td::actor::Actor {
                                                                   .public_broadcast_speed_multiplier_ = 3.33,
                                                                   .private_broadcast_speed_multiplier_ = 3.33,
                                                                   .fast_sync_broadcast_speed_multiplier_ = 3.33,
-                                                                  .initial_sync_delay_ = 60.0};
+                                                                  .initial_sync_delay_ = 0.0};
 
   std::map<ton::BlockSeqno, std::pair<ton::CatchainSeqno, td::uint32>> unsafe_catchain_rotations_;
   ton::quic::QuicServer::Options quic_options_ = {};
--- a/validator/full-node-custom-overlays.cpp
+++ b/validator/full-node-custom-overlays.cpp
@@ -492,6 +492,24 @@ void FullNodeCustomOverlay::download_block(BlockIdExt id, td::uint32 priority, t
       .release();
 }
 
+void FullNodeCustomOverlay::download_next_block(BlockIdExt prev_id, td::uint32 priority, td::Timestamp timeout,
+                                                td::Promise<ReceivedBlock> promise) {
+  if (!inited_) {
+    promise.set_error(td::Status::Error(ErrorCode::notready, "custom overlay is not ready"));
+    return;
+  }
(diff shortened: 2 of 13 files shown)


validator-engine/validator-engine.hpp
```diff
@@ -278,7 +278,7 @@ class ValidatorEngine : public td::actor::Actor {
                                                                   .public_broadcast_speed_multiplier_ = 3.33,
                                                                   .private_broadcast_speed_multiplier_ = 3.33,
                                                                   .fast_sync_broadcast_speed_multiplier_ = 3.33,
-                                                                  .initial_sync_delay_ = 60.0};
+                                                                  .initial_sync_delay_ = 0.0};
```
DanShaders, line 281: This is stray.

SpyCheese:
See #2442

PR #2432, opened 2026-06-09
http: reject CR/LF in request url and response reason phrase

kesha-antonov:
## What

Reject CR/LF in the HTTP request-line `url` and the response `reason` phrase at construction time.

## Why

`HttpRequest::store_http` / `HttpResponse::store_http` write `url` and `reason` verbatim into the first line of the serialized message (`METHOD url PROTO\r\n` and `PROTO code reason\r\n`). Header names and values are already checked for CR/LF in `HttpHeader::basic_check()`, but `url` and `reason` are not validated anywhere.

This matters because `rldp-http-proxy` rebuilds an `HttpRequest`/`HttpResponse` directly from the TL fields (`http_request.url` / `http_response.reason`) and re-serializes it onto a downstream plain-HTTP connection without ever re-parsing. A CR/LF in those upstream-supplied fields therefore injects extra request/response lines into the downstream connection - request smuggling on the gateway side, response splitting on the client side. The headers can't be abused this way because `basic_check()` already rejects CR/LF there; the request-line `url` and the status-line `reason` are the only unchecked fields on the first line.

## Change

- `HttpRequest::create` rejects CR/LF in `url`
- `HttpResponse::create` rejects CR/LF in `reason`
- both checks live in `create()` so every construction path (the line parser and the proxy's TL-based reconstruction) is covered, mirroring the existing header check
- test added in `test/test-http.cpp`

Legitimate traffic is unaffected: a url or reason phrase never contains a raw CR/LF, and the line parser strips the line terminator before calling `create()`.

### DIFF
--- a/http/http.cpp
+++ b/http/http.cpp
@@ -146,6 +146,16 @@ td::Result<std::unique_ptr<HttpRequest>> HttpRequest::create(std::string method,
     return td::Status::Error(PSTRING() << "unsupported http method '" << method << "'");
   }
 
+  // The request line is later re-serialized verbatim ("method url proto\r\n").
+  // CR/LF in the url would let an upstream-supplied value inject extra request
+  // lines into the downstream connection (request smuggling), so reject them
+  // here - the same way HttpHeader::basic_check() guards header values.
+  for (const auto &c : url) {
+    if (c == '\r' || c == '\n') {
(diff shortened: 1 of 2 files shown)


SpyCheese:
We already have a fix for this planned for the next update.

PR #2434, opened 2026-06-13
chore: remove cache poisoning

Danil42Russia:


DanShaders (changes requested):
I would actually prefer if we drop this block altogether, CMAKE_* variables are not ours to set.

neodix42 (approved):

DanShaders (approved):

PR #2441, opened 2026-06-18
Tolk v1.4.2: rename TON to GRAM

tolk-vm:
This is a minor update, mostly required to reflect step 4 of MTONGA.

## Rename `ton("0.05")` to `grams("0.05")`

The currency name is being changed from "TON" to "GRAM". Hence, this renaming has been done throughout the compiler, stdlib, and doc comments.

For instance, a built-in function `ton` is becoming deprecated: use `grams`

```
// before
val amount = ton("0.05");     // 50000000 nanotons

// now
val amount = grams("0.05");   // 50000000 nanograms
```

**The old `ton` function is still available for backward compatibility**. So, rewriting your code immediately is not required, though recommended for consistency.

Similarly, `reserveToncoinsOnBalance` becomes deprecated in favor of `reserveGramsOnBalance`.

Note that many standard contracts have field naming like `forwardTonAmount`. Probably, you have similar in your contracts. They can be safely renamed to `forwardGrams` etc., because field names do not affect binary encoding.

## Minor changes in a language

Besides, there are small changes and bugfixes in corner cases.

1. **Tensors are not valid lvalues anymore**. Example: `f(mutate (a,b))` and `(a, b).0 = newA` are now forbidden. This is absolutely useless in practice, but could lead to unexpected side effects.
  Note that syntax `(a, b) = rhs` is allowed. It's now treated as a special destructuring syntax, as well as `[a, b] = rhs`. In the future, destructuring of structs will be supported the same way.

2. **Disallow `SomeAlias { ... }`, allow only `SomeStruct { ... }`**. This restriction was introduced to prevent ambiguilty when structs/aliases contradict in custom serializers. 

3. **Disallow `lazyVar = rhs`, allow only `lazyVar.field = rhs`**. Whole-lazy assignments are impractical, but may lead to corner cases bugs like `(v = v).toCell()`.

4. **Minor bug fixes of issues found by LLM fuzzing**. Traditionally, we continuously scan the compiler to find corner cases where it mismatches specification or crashes. All findings are absolutely impractical.
For example, recursive implicit generics that appear in constant initializers.
For example, assigning one `lazy` variable to another, which probably should be prohibited. 
For example, `assert (condition) throw funThatThrows()`.
And so on.

## Related pull requests

* https://github.com/ton-blockchain/tolk-js/pull/21
* https://github.com/ton-org/docs/pull/2222
* https://github.com/ton-blockchain/acton/pull/1128
* https://github.com/ton-blockchain/acton/pull/1139

### DIFF
--- a/crypto/smartcont/tolk-stdlib/common.tolk
+++ b/crypto/smartcont/tolk-stdlib/common.tolk
@@ -68,8 +68,8 @@ type intN = builtin
 /// overflow will happen at serialization to a cell/builder, NOT at assignment. 
 type uintN = builtin
 
-/// `coins` is a special primitive representing "nanotoncoins". One TON = 10^9 nanotoncoins.
-/// You can create coins with `ton()` function: `ton("0.05")` (actually, `int` 50000000 at runtime).
+/// `coins` is a special primitive representing "nanograms". One GRAM = 10^9 nanograms.
+/// You can create coins with `grams()` function: `grams("0.05")` (actually, `int` 50000000 at runtime).
 /// Arithmetic operations on `coins` degrade to 257-bit `int` type.
 type coins = builtin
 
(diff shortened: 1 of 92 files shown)


crypto/smartcont/tolk-stdlib/common.tolk
```diff
@@ -68,8 +68,8 @@ type intN = builtin
 /// overflow will happen at serialization to a cell/builder, NOT at assignment. 
 type uintN = builtin
 
-/// `coins` is a special primitive representing "nanotoncoins". One TON = 10^9 nanotoncoins.
-/// You can create coins with `ton()` function: `ton("0.05")` (actually, `int` 50000000 at runtime).
+/// `coins` is a special primitive representing "nanograms". One GRAM = 10^9 nanograms.
+/// You can create coins with `grams()` function: `grams("0.05")` (actually, `int` 50000000 at runtime).
```
aleksola, line 72: Grams
comment_payer_hl.tolk

crypto/smartcont/tolk-stdlib/common.tolk
```diff
@@ -68,8 +68,8 @@ type intN = builtin
 /// overflow will happen at serialization to a cell/builder, NOT at assignment. 
 type uintN = builtin
 
-/// `coins` is a special primitive representing "nanotoncoins". One TON = 10^9 nanotoncoins.
-/// You can create coins with `ton()` function: `ton("0.05")` (actually, `int` 50000000 at runtime).
```
aleksola: Ton

PR #2477, opened 2026-07-15
Cleanup consensus and broadcast code

SpyCheese:
- Remove code for proto version < 2
- Remove old block broadcasts
- Improve candidate and finality processing in manager.cpp


Mustang98 (approved):
OK

PR #2487, opened 2026-07-24
Fix storing is_applied flag

SpyCheese:


validator/state-serializer.cpp
```diff
@@ -156,13 +156,26 @@ void AsyncStateSerializer::next_iteration() {
   }
   if (!masterchain_handle_->inited_unix_time() || !masterchain_handle_->inited_is_key_block() ||
       !masterchain_handle_->is_applied()) {
+    if (last_known_key_block_.is_valid() && masterchain_handle_->inited_unix_time() &&
```
AndreyMlashkin, line 159: Please handle the no-timestamp case using information that does not come from the corrupted handle.

This change makes recovery unreachable for a missing handle. The lookup uses get_block_handle(..., true), which substitutes an empty handle when the DB returns notready; that handle has neither inited_unix_time() nor inited_next_left(). The serializer therefore returns here without recovering, and the next top-block check can abort on CHECK(masterchain_handle_->inited_next_left()).

validator/downloaders/download-state.cpp
```diff
@@ -491,6 +491,7 @@ void DownloadShardState::written_shard_state(td::Ref<ShardState> state) {
   auto P = td::PromiseCreator::lambda([SelfId = actor_id(this), handle = handle_](td::Result<td::Unit> R) {
     CHECK(handle->handle_moved_to_archive());
     CHECK(handle->moved_to_archive())
+    handle->set_applied_stored();
```
AndreyMlashkin: Please check R first, explicitly flush/update dirty already-moved handles, and set applied_stored only after that write succeeds.

archive() does not guarantee that the changes made above were persisted. BlockArchiver::run_inner() returns success immediately when moved_to_archive() is already true, so set_applied() can merely dirty an existing archived handle while this line nevertheless marks it as stored. ApplyBlock can then take the new early-return path even though the archive still contains is_applied=false.

validator/state-serializer.cpp
SpyCheese: This path is intended for handles with missing applied flag, which is technically possible and was actually observed on some nodes. We don't expect corrupted handles with other missing flags.

validator/downloaders/download-state.cpp
SpyCheese: Fixed

birydrad (approved):

PR #2566, opened 2026-09-15
Collation optimizations: cache successful Ed25519 signature verifications

Ormlis:
- Add bounded concurrent seqlock cache 
- Current cache size is 65536 slots (15 MiB), but can be changed.
- Add the `Ed25519_verify_signature_cache_hit` performance counter

Benchmark (1M accounts, 128 jetton transfers, 32 threads; 3 runs x 100 iterations, 5 warmups):

| Branch | Commit | Collate | Validate (collated) | Validate (preloaded) | Δ Collate |
| --- | --- | ---: | ---: | ---: | ---: |
| testnet (baseline) | `8534fc7d6268` | 83.150 | 36.983 | 30.525 | 0.00% |
| collation-opts-ed25519-cache | `fe5097237e1b` | 72.623 | 37.180 | 31.059 | -12.66% |


crypto/vm/tonops.cpp
```diff
@@ -787,17 +829,39 @@ int exec_ed25519_check_signature(VmState* st, bool from_slice) {
...
+  auto* cache = get_signature_cache();
+  if (!cache) {
+    stack.push_bool(verify_signature() || st->get_chksig_always_succeed());
+    return 0;
+  }
+  auto cache_key = make_signature_cache_key(key, signature, data, data_len);
+  auto slot = cache->key_to_slot(cache_key);
+  bool signature_ok = cache->contains(cache_key, slot);
+  if (signature_ok) {
+    TD_PERF_COUNTER(VM_check_signature_cache_hit);
```
Ormlis, line 853: It might be worth adding a counter primitive that only counts events, without tracking duration, to avoid the timing overhead of TD_PERF_COUNTER

PR #2568, opened 2026-09-16
Collation optimizations: augmented dictionary batching

Ormlis:
- Add multiset support for AugmentedDictionary to reduce repeated trie traversal and cell rebuilding
- Reuse the account_dict_estimator and batch other updates during finalization
- Batch In/OutMsgDescr insertions in groups of 64

Benchmark (1M accounts, 128 jetton transfers, 32 threads; 3 runs x 100 iterations, 5 warmups):

| Branch | Commit | Collate | Validate (collated) | Validate (preloaded) | Δ Collate |
| --- | --- | ---: | ---: | ---: | ---: |
| testnet (baseline) | `8534fc7d6268` | 83.150 | 36.983 | 30.525 | 0.00% |
| collation-opts-ed25519-cache | `fe5097237e1b` | 72.623 | 37.180 | 31.059 | -12.66% |
| collation-opts-dict-multiset | `4d98337fcb24` | 72.099 | 36.116 | 30.882 | -13.29% |
| cache + dict (merged) | - | 61.149 | 36.596 | 30.604 | -26.46% |


Lordron:
Can we benefit of using a small-buffer-optimized container for instance in flush_out_msg_descriptors ?

Ormlis:
> Can we benefit of using a small-buffer-optimized container for instance in flush_out_msg_descriptors ?

It could help, but I'd expect the gain to be small. Most of the time is likely spent in multiset, so I don't think we need this optimization for now

